PR #3, opened 2024-11-12
Add openssf scorecard scan

huiwangredhat:
I found the [Scorecard](https://github.com/ossf/scorecard?tab=readme-ov-file#scorecard-github-action) provides the GitHub Action which is the easiest way to use it.
So I copied it and did a minor update for the action trigger.

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,73 @@
+# This workflow uses actions that are not certified by GitHub. They are provided
+# by a third-party and are governed by separate terms of service, privacy
+# policy, and support documentation.
+
+name: Scorecard supply-chain security
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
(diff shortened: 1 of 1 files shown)


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,72 @@
...
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  push:
+    branches: [ "main" ]
+  pull_request:
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
+  analysis:
```
jpower432, line 20: I recommend adding an `if` statement here to check which repository it is running in to make sure it does not run on pushes to forks.

jpower432 (commented):
Looks good. One suggestion.

.github/workflows/scorecard.yml
huiwangredhat: Good point! I added the `if` statement.

jpower432 (approved):
LGTM

PR #5, opened 2024-11-19
Add go lint

huiwangredhat:
This is the simple use of [golangci-lint-action](https://github.com/golangci/golangci-lint-action).

The enabled by default linters as follows: 
errcheck, gosimple, govet, ineffassign, staticcheck, unused.

I propose to add more linters from the list of [Disabled by default](https://golangci-lint.run/usage/linters/#disabled-by-default).

Your thoughts about which linters should be added is always appreciated.

### DIFF
--- a/.github/workflows/go_lint.yml
+++ b/.github/workflows/go_lint.yml
@@ -1,0 +1,23 @@
+name: golangci-lint
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5
+        with:
+          go-version-file: './go.mod'
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@971e284b6050e8a5849b72094c50ab08da042db8 #v6
+        with:
+          version: v1.60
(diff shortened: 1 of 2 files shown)


jpower432 (commented):
@huiwangredhat I recommend adding `gosec` and `goimports` to the list. ~~Maybe we should think about `gocyclo` as well~~. What do you think?

**Edited**: Actually I'd imagine we would want to measure cyclomatic complexity elsewhere and not enforce it here.

huiwangredhat:
@jpower432 

> @huiwangredhat I recommend adding `gosec` and `goimports` to the list. ~Maybe we should think about `gocyclo` as well~. What do you think?
> 
> **Edited**: Actually I'd imagine we would want to measure cyclomatic complexity elsewhere and not enforce it here.

Thanks for your feedback. I agree to add `gosec` to linter which is focused on the potential secerity issue.                               

> I have some concerns about adding `goimports` to linter. Because it could update the import part code automatically. It will cause conflicts between the code that passes the CI checks and the code locally committed. 
> I propose to add the requirement to the CONTRIBUTING document.
> -  Run goimports Locally: Developers run `goimports -w .` on their local machines to format code before pushing it to version control.
> -  Use Pre-commit Hooks: Implement a pre-commit hook that automatically runs goimports before every commit, ensuring that only properly formatted code enters the repository.
> 
Ok, I installed the `goimports` locally and did some research. The` goimports -w` could update the import code part. It won't update the code if no the parameter `-w`. So I think we could include `goimports` to linters list.  For the `gocyclo`, it will give the value of code complexity in the CI, no error or failed. I am ok to add it or not add it.

jpower432:
@huiwangredhat Everything looks good here. Based on the alerts, it looks like the specified actions need to be pinned to a commit hash.

marcusburghardt (approved):
LGTM. There are some interesting linters disabled by default but I am not sure the value they would bring at this moment. We can include more in the future, if necessary. The proposed ones seem pretty reasonable. Thanks.

jpower432 (approved):
LGTM

PR #7, opened 2024-11-22
CPLYTM-341 CPLYTM-354 Add initial goreleaser config

d10n:


.goreleaser.yaml
```diff
@@ -0,0 +1,84 @@
...
+#      - darwin
+  - #
+    id: linux
+    env:
+      - CGO_ENABLED=0
+    goos:
+      - linux
+    ldflags:
+      - "-X github.com/complytime/complytime/version.version={{ .Version }}"
+      - "-X github.com/complytime/complytime/version.gitTreeState={{ .Version }}"
```
marcusburghardt: Using `.Version` here would not be aligned to the Makefile content. Can we make them consistent?

.goreleaser.yaml
```diff
@@ -0,0 +1,84 @@
...
+      - "-X github.com/complytime/complytime/version.commit={{ .Commit }}"
+      - "-X github.com/complytime/complytime/version.buildDate={{ .Date }}"
+      - "-X github.com/complytime/complytime/version.commitDate={{ .CommitDate }}"
```
marcusburghardt: `commitDate` is not present in the Makefile content. Can we make them consistent?

.goreleaser.yaml
```diff
@@ -0,0 +1,84 @@
...
+#  GO_LD_EXTRAFLAGS :=-X github.com/complytime/complytime/version.version="$(GIT_TAG)" \
```
marcusburghardt: I assume the commented lines like this are used for reference for now and will be removed, correct?

.goreleaser.yaml
```diff
@@ -0,0 +1,84 @@
...
+    name_template: >-
+      {{ .ProjectName }}_
+      {{- title .Os }}_
+      {{- if eq .Arch "amd64" }}x86_64
+      {{- else if eq .Arch "386" }}i386
+      {{- else }}{{ .Arch }}{{ end }}
+      {{- if .Arm }}v{{ .Arm }}{{ end }}
+    # use zip for windows archives
+    format_overrides:
+      - goos: windows
```
marcusburghardt: Do we need this since it will build only for Linux?

.goreleaser.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+changelog:
+  sort: asc
+  filters:
+    exclude:
+      - "^docs:"
+      - "^test:"
+
+
+nfpms:
```
marcusburghardt: I wonder if it is a good idea to provide a RPM in Upstream. In general, each distro provides their own RPMs though their own processes in order to make them available for the users. So I imagine a RPM here would be more for testing purposes. Is that the intention?

marcusburghardt (commented):
I read the GoReleaser documentation but I don't have experience releasing with it, so take this into consideration when reading my comments, please.

.goreleaser.yaml
```diff
@@ -0,0 +1,84 @@
...
+  - #
+    id: rpm
+    package_name: complytime
+    builds:
+      - linux
+    if: '{{ eq .Os "linux" }}'
+    vendor: Red Hat
+    homepage: https://github.com/complytime/complytime
+    maintainer: ComplyTime Maintainers <complytime@redhat.com>
```
huiwangredhat: Did we apply the email? Or here is it just a placeholder?

.goreleaser.yaml
huiwangredhat: It could provide a noarch RPM that could be available for all the distributions, right? I think the 
 `complytime` rpm should be used in different arch distributions. Am I right?

.goreleaser.yaml
jpower432: GoReleaser needs a [clean state to work](https://goreleaser.com/errors/dirty/). I think the `gitTreeState` could just be set to clean.

.goreleaser.yaml
d10n: We do have the mailing list

.goreleaser.yaml
d10n: Removed

.goreleaser.yaml
d10n: I had intended to keep them for convenience of inline reference but I'll remove them

.goreleaser.yaml
d10n: Copy-paste error!

.goreleaser.yaml
d10n: It's not used in version.go but I thought it might be useful to consider adding. I'll remove it.

.goreleaser.yaml
d10n: This was a copy-paste error. GoReleaser does have a .GitTreeState variable and I assume it could be dirty when building only (ie not releasing), so I updated it to use that

huiwangredhat (dismissed):
LGTM.

.goreleaser.yaml
```diff
@@ -0,0 +1,48 @@
...
+version: 2
+
+before:
+  hooks:
+    - go mod tidy
+    - go mod vendor
+    #- go generate ./...
+
+builds:
+#  - env:
```
marcusburghardt: It is probably a nit, but since `env` and `goos` are already used below, these comments seem unnecessary.

.goreleaser.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - "-X github.com/complytime/complytime/version.commit={{ .Commit }}"
+      - "-X github.com/complytime/complytime/version.buildDate={{ .Date }}"
+    main: ./cmd/complytime/
+
+archives:
+  - format: tar.gz
+    # this name template makes the OS and Arch compatible with the results of `uname`.
+    name_template: >-
+      {{ .ProjectName }}_
+      {{- title .Os }}_
```
marcusburghardt:
```suggestion
      {{- .Os }}_
```
What do you think about removing `Title` so we have final files named, for example, as `complytime_linux_x86_64.tar.gz` instead of `complytime_Linux_x86_64.tar.gz`?

marcusburghardt (commented):
Thanks for the updates. I have only two comments for your consideration.

.goreleaser.yaml
d10n: I'll remove it, I left it in for reference but it's not too important

.goreleaser.yaml
```diff
@@ -0,0 +1,44 @@
...
+    goos:
+      - linux
+    ldflags:
+      - "-X github.com/complytime/complytime/version.version={{ .Version }}"
+      - "-X github.com/complytime/complytime/version.gitTreeState={{ .GitTreeState }}"
+      - "-X github.com/complytime/complytime/version.commit={{ .Commit }}"
+      - "-X github.com/complytime/complytime/version.buildDate={{ .Date }}"
+    main: ./cmd/complytime/
+
+archives:
```
jpower432, line 28: Not a blocking comment, but adding checksums would be good https://goreleaser.com/customization/checksum/

jpower432 (approved):
LGTM, one non-blocking comment.

marcusburghardt (approved):
Thanks for the updates on this @d10n .

PR #8, opened 2024-11-22
CPLYTM-309 Add initial pre-commit config

d10n:
Use pre-push so that individual commits are allowed to fail unit tests, etc.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,19 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+default_stages: [pre-push]
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v5.0.0
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-yaml
+    -   id: check-added-large-files
+-   repo: https://github.com/dnephin/pre-commit-golang
+    rev: v0.5.1
+    hooks:
+      - id: go-fmt
+      #- id: go-imports
+      #- id: no-go-testing
+      #- id: golangci-lint
+      - id: go-unit-tests
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,14 +3,14 @@
 default_stages: [pre-push]
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v5.0.0
+    rev: cef0300fd0fc4d2a87a85fa2093c6b283ea36f4b #v5.0.0
     hooks:
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
     -   id: check-yaml
     -   id: check-added-large-files
 -   repo: https://github.com/dnephin/pre-commit-golang
-    rev: v0.5.1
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,19 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+default_stages: [pre-push]
```
huiwangredhat, line 3: I met one issue that's interesting. 
**If this default_stages is set, the hooks will be run as follows:**
```
% git commit --amend 
trim trailing whitespace.........................Passed
fix end of files........................................Passed
check for added large files......................Passed
```
**If I comment on this set, it will be run as follows:**
```
git commit --amend
trim trailing whitespace........................................Passed
fix end of files.........................................................Passed
check yaml.............................................................Passed
check for added large files....................................Passed
go fmt...............................................Passed
go-unit-tests........................................Passed
```

.pre-commit-config.yaml
```diff
@@ -0,0 +1,19 @@
...
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v5.0.0
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-yaml
+    -   id: check-added-large-files
```
marcusburghardt, line 11: The [default size is 500k](https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#check-added-large-files), which I believe is a generous number for code files. Should we specify a lower number? Or is it the intention to catch other types of files, such as images, binaries, etc?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,19 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+default_stages: [pre-push]
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v5.0.0
```
marcusburghardt: Could you pin the versions to the commit hashes (e.g.: [scorecard.yml](https://github.com/complytime/complytime/blob/main/.github/workflows/scorecard.yml#L35)) instead?

More context: https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

.pre-commit-config.yaml
d10n: I chose this on purpose because if a person wants to commit something while developing or testing and the tests fail, it may discourage committing during the development process. Unit tests should pass before pushing updates, though, so all the hooks run when running `git push`

.pre-commit-config.yaml
d10n: 500k seems reasonable to me. It is comfortably larger than any source code file, and probably large enough for test data files, but smaller than binary files that would be likely accidentally added, but I don't have a strong opinion.

huiwangredhat (approved):
LGTM.

.pre-commit-config.yaml
marcusburghardt: Ok. Thanks for the clarification. Along the time we may have a better idea of sizes but I think it is fine for now.

marcusburghardt (approved):
LGTM. Thanks for including this @d10n

PR #11, opened 2024-12-04
CPLYTM-342 Add initial release workflows

d10n:
To be merged after #7

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -5,16 +5,14 @@ on:
     branches:
       - main
 
-permissions:
-  contents: write
-  # packages: write
-  # issues: write
-  # id-token: write
+permissions: {}
 
 jobs:
   release:
     name: Create Pre-release
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
 
     steps:
 
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -19,7 +19,7 @@ jobs:
     steps:
 
       # Prepare token
-      - uses: actions/create-github-app-token@v1
+      - uses: actions/create-github-app-token@5d869da34e18e7287c1daad50e0b8ea0f506ce69 #v1
         id: app-token
         with:
           # required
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,80 @@
...
+jobs:
+  release:
+    name: Create Pre-release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+
+      # Prepare token
```
jpower432: I am wondering if it would be simpler here to either use the GitHub App token for subsequent steps or configure the GitHub Actions bot as the git user. That way we would not have to generate another token just to get the user id. What you think?

For the Actions bot you can just [do](https://github.com/actions/checkout?tab=readme-ov-file#push-a-commit-using-the-built-in-token)
```bash
 git config user.name "github-actions[bot]"
 git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```

d10n: I believe that the token is used for `git push origin "$new_tag"` below

jpower432: I think you would have to clone the repository on line 35 with the GitHub App token for it to be used in the git push. An example of what I am referring to is [here](https://github.com/actions/create-github-app-token?tab=readme-ov-file#use-app-token-with-actionscheckout).

d10n:
Rebased

.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
+name: Full Release
+
+on:
+  workflow_dispatch:
+  #push:
```
jpower432, line 5: Was this intentionally left commented?

jpower432 (approved):
LGTM with one question

gvauter (approved):
LGTM - same question as Jenn on the commented lines in release.yml

.github/workflows/release.yml
d10n: I did leave it commented on purpose, for future reference if this manual trigger becomes too cumbersome. I think I'd like to keep it for now, but we can remove it if we like the workflow

PR #13, opened 2024-12-06
CPLYTM-376 CPLYTM-343 Add terminal spinner function

d10n:


hbraswelrh (commented):
LGTM

pkg/utils.go
```diff
@@ -0,0 +1,28 @@
+package pkg
```
jpower432: Since `pkg` by convention is used to store other packages, I think importing from here would unexpected. If this is intended to be a top-level package like `version` is, perhaps this could be name to something else to improve the clarity like `progress`.

pkg/utils.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"time"
+)
+
+var out io.Writer = os.Stdout
```
jpower432: Nit: I see that the `out` variable is being set to something more easily testable during unit testing. I think passing this in as an argument to `ShowSpinner` instead of setting it at the package level makes the dependency more explicit and reduces the chance of it being modified somewhere else in the package.

jpower432 (changes requested):
Added a couple suggestions here. One of them is just a nit.

pkg/utils.go
d10n: I weakly disagree but I'll make the change.
I originally considered making this an argument, but it would always be stdout except for tests, so it would be annoying, and I made a choice between annoyance with full test coverage, a helper function with less test coverage, and no annoyance with full test coverage

jpower432: Thanks @d10n. It might be worth waiting for another review so see if other folks might disagree with me as well. This is non-blocking comment from me.

jpower432 (approved):
LGTM

hbraswelrh (approved):
LGTM

d10n:
We're looking at TUI libraries and some have this built-in. Let's merge this for now and remove it upon picking one

PR #18, opened 2024-12-20
Add spdx license hook

d10n:
It works but pre-commit calls the hook twice sometimes and the output gets doubled.


d10n:
Closing in favor of #73

PR #20, opened 2024-12-23
Add zap/hclog interface

d10n:


d10n:
Accidentally targeted a branch other than main and then deleted it, reopened as #21

PR #29, opened 2025-01-20
CPLYTM-476 - xccdf package for openscap-plugin

marcusburghardt:
## Summary

With openscap-plugins the system will be scanned with the Datastream files provided by `scap-security-guide` package.
However, the OSCAL profile may select different rules and parameters from the Datastream profile.
In this scenario, the openscap-plugin will use a tailoring file based on the OSCAL Profile to scan the system.

This PR introduces the `xccdf` package for openscap-plugin to create the initial structure to manage OpenSCAP Tailoring files. The existing code in `server.go` that was related to tailoring and XCCDF was also moved to the new package.

## Related Issues

- Relates to CPLYTM-476

## Review Hints

Take a look at the commit descriptions for more details.
Also take into consideration that more code is expected for this package, so this PR includes the minimal working structure to enable more changes in separate PRs.

For unit tests
`make test-unit`

### DIFF
--- a/cmd/openscap-plugin/xccdf/tailoring_test.go
+++ b/cmd/openscap-plugin/xccdf/tailoring_test.go
@@ -1,0 +1,321 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package xccdf
+
+import (
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/ComplianceAsCode/compliance-operator/pkg/xccdf"
+	"github.com/complytime/complytime/cmd/openscap-plugin/config"
+	"github.com/oscal-compass/compliance-to-policy-go/v2/policy"
+	"github.com/oscal-compass/oscal-sdk-go/extensions"
+)
+
+// TestGetTailoringID tests the getTailoringID function.
+func TestGetTailoringID(t *testing.T) {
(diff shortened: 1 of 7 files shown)


cmd/openscap-plugin/xccdf/tailoring_test.go
```diff
@@ -0,0 +1,321 @@
...
+					ID:          "param1",
+					Description: "param1_description",
+					Value:       "value1",
+				},
+			},
+		},
+	}
+
+	cfg := &config.Config{
+		Files: struct {
```
jpower432: Nit: This might be an isolated usage of this type, if so, please disregard. If using the `config.Config` type, it may be more readable to use a named struct in the `config.Config` type instead of anonymous struct so you do not have to redefine the struct fields each time.

jpower432 (dismissed):
LGTM. Added one nit.

cmd/openscap-plugin/xccdf/tailoring_test.go
marcusburghardt: Your point is pretty valid @jpower432 . Thanks for the hint. I just sent an additional commit addressing this. Could you take a look at the change, please?

jpower432 (approved):
LGTM

gvauter (approved):
LGTM

PR #33, opened 2025-01-27
CPLYTM-474 - Allow openscap-plugin to collect variables and options from Datastream

marcusburghardt:
## Summary
As part of the tailoring file creation the plugin will collect all variables and respective values in Datastream and use this information to create a minimalist Tailoring file that includes only the differences between OSCAL profile and Datastream profile.

The logic to compare and report the differences is not in the scope of this PR and expected by separate PRs.

In Datastreams the profiles include variables and respective options relevant for each profile, but not the actual values for these options. The values are defined in the scope of a variable definition in Datastream. Since OSCAL is expected to work directly with values instead of options, this PR also introduces logic to resolve the Datastream profile options to actual values in order much easier compare the profiles later.

## Related Issues

- Relates to CPLYTM-474

## Review Hints

The commit descriptions should provide the context and more details about each change.
To make the code more readable, some aesthetic changes and refactoring were made along the commits.
Therefore, while the commit descriptions give context and details to clarify how the PR evolved, it could be easier to analyse the final result.

For unit tests
`make test-unit`

### DIFF
--- a/cmd/openscap-plugin/xccdf/datastream.go
+++ b/cmd/openscap-plugin/xccdf/datastream.go
@@ -10,15 +10,6 @@ import (
 	"github.com/antchfx/xmlquery"
 )
 
-func getDsProfileID(profileId string) string {
-	return fmt.Sprintf("xccdf_org.ssgproject.content_profile_%s", profileId)
-}
-
-func getDsElement(dsDom *xmlquery.Node, dsElement string) (*xmlquery.Node, error) {
-	// Returns nil if the element is not found
-	return xmlquery.Query(dsDom, dsElement)
-}
-
 func loadDataStream(dsPath string) (*xmlquery.Node, error) {
 	file, err := os.Open(dsPath)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


hbraswelrh (approved):
LGTM

cmd/openscap-plugin/xccdf/datastream.go
```diff
@@ -33,27 +38,260 @@ func loadDataStream(dsPath string) (*xmlquery.Node, error) {
...
+func populateProfileInfo(dsProfile *xmlquery.Node, parsedProfile *xccdf.ProfileElement) (*xccdf.ProfileElement, error) {
+	profileTitle, err := getDsElementTitle(dsProfile)
+	if err != nil {
+		return parsedProfile, fmt.Errorf("error populating profile title: %s", err)
+	}
+	if parsedProfile.Title == nil {
+		parsedProfile.Title = &xccdf.TitleOrDescriptionElement{}
+	}
+	if profileTitle == nil {
+		// log that profile title was not found.
```
gvauter, line 106: Is this a note to add logging in the future or did you intend to include here?

gvauter (approved):
LGTM

cmd/openscap-plugin/xccdf/datastream.go
marcusburghardt: My plan was to include in the future. I need to take a better look and do some tests with [log.go](https://github.com/complytime/complytime/blob/main/pkg/log/log.go)

PR #35, opened 2025-01-28
CPLYTM-473: collect Datastream rules

hbraswelrh:
## Summary

The implementation of `populateProfileRules` function is to extract selected rules from the OpenSCAP Datastream file. The selections are initialized in the `initProfile` command. The error handling will ensure that populated profile rules are accessible through the profile elements "idref" and their associated "selected" field is included from the OpenSCAP Datastream file. The `TestpopulateProfileRules` serves as the unit test associated with `populateProfileRules`.

## Review Hints

- The two commits included in this PR have descriptions that highlight their implementations. The only new introductions for this PR are the `populateProfileRules` function and `TestpopulateProfileRules` unit test.

### DIFF
--- a/cmd/openscap-plugin/xccdf/datastream.go
+++ b/cmd/openscap-plugin/xccdf/datastream.go
@@ -3,9 +3,9 @@
 package xccdf
 
 import (
-	"encoding/xml"
 	"fmt"
 	"os"
+	"strconv"
 
 	"github.com/ComplianceAsCode/compliance-operator/pkg/xccdf"
 	"github.com/antchfx/xmlquery"
@@ -22,12 +22,6 @@ type DsVariables struct {
 	Title       string `xml:",chardata"`
 	Description string `xml:",chardata"`
 	Options     []DsVariableOptions
-}
-
-type SelectElement struct {
-	XMLName  xml.Name `xml:"xccdf-1.2:select"`
-	IDRef    string   `xml:"idref,attr"`
-	Selected bool     `xml:"selected,attr"`
 }
 
 func loadDataStream(dsPath string) (*xmlquery.Node, error) {
@@ -171,28 +165,34 @@ func populateProfileRules(dsProfile *xmlquery.Node, parsedProfile *xccdf.Profile
 	if parsedProfile.Selections == nil {
 		parsedProfile.Selections = []xccdf.SelectElement{}
 	}
+
 	profileRules, err := getDsElements(dsProfile, "xccdf-1.2:select")
 	if err != nil {
 		return parsedProfile, fmt.Errorf("error finding 'select' elements in profile: %s", err)
 	}
+
 	for _, rule := range profileRules {
 		ruleIdRef, err := getDsElementAttrValue(rule, "idref")
 		if err != nil {
 			return parsedProfile, fmt.Errorf("error getting value of 'idref' attribute: %s", err)
 		}
-		// change occurrences of ruleSelector to ruleSelected for alignment with selected attribute
 		ruleSelected, err := getDsElementAttrValue(rule, "selected")
 		if err != nil {
-			return parsedProfile, fmt.Errorf("error getting value of 'selected' attribute: %s", err)
+			return nil, fmt.Errorf("error getting value of 'selected' attribute: %s", err)
+		}
+		selectedBoolean, err := strconv.ParseBool(ruleSelected)
+		if err != nil {
+			return nil, fmt.Errorf("error converting the 'selected' attribute from string to boolean: %s", err)
 		}
 
 		parsedProfile.Selections = append(parsedProfile.Selections, xccdf.SelectElement{
 			IDRef:    ruleIdRef,
-			Selected: ruleSelected,
+			Selected: selectedBoolean,
 		})
 	}
 	return parsedProfile, nil
 }
+
 func initProfile(dsProfile *xmlquery.Node, dsProfileId string) (*xccdf.ProfileElement, error) {
 	parsedProfile := new(xccdf.ProfileElement)
 	parsedProfile.ID = dsProfileId
@@ -201,9 +201,6 @@ func initProfile(dsProfile *xmlquery.Node, dsProfileId string) (*xccdf.ProfileEl
 	if err != nil {
 		return parsedProfile, fmt.Errorf("error populating profile title and description: %s", err)
 	}
-
-	// Here we can add the logic to populate profile rules in a separate PR
-	// TODO: Implement the profile rules logic - to extract the Selected
 
 	parsedProfile, err = populateProfileRules(dsProfile, parsedProfile)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/openscap-plugin/xccdf/datastream_test.go
```diff
@@ -691,6 +723,37 @@ func TestResolveDsVariableOptions(t *testing.T) {
...
 
+// TestGetDsSelected tests the GetDsProfileRules function.
+func TestGetDsProfileRules(t *testing.T) {
+	tests := []struct {
+		dsPath   string
+		expected int
+		wantErr  bool
+	}{
+		// Declaration of the structure types for test data
+		{filepath.Join(testDataDir, "ssg-rhel-ds.xml"), 71, false},
```
marcusburghardt:
```suggestion
		{filepath.Join(testDataDir, "ssg-rhel-ds.xml"), 376, false},
```
The Datastream used for tests contains 71 variables but 376 rules. But we could also remove this unit test since the function to be tested can also be removed in this PR.

cmd/openscap-plugin/xccdf/datastream.go
```diff
@@ -23,6 +24,13 @@ type DsVariables struct {
 	Options     []DsVariableOptions
 }
 
+// Existing struct in compliance-operator compliance-operator/blob/master/pkg/xccdf/tailoring.go
```
marcusburghardt: We can remove this commented struct probably used during the tests. : )

cmd/openscap-plugin/xccdf/datastream.go
```diff
@@ -160,6 +168,36 @@ func populateProfileVariables(dsProfile *xmlquery.Node, parsedProfile *xccdf.Pro
...
+	profileRules, err := getDsElements(dsProfile, "xccdf-1.2:select")
+	if err != nil {
+		return parsedProfile, fmt.Errorf("error finding 'select' elements in profile: %s", err)
+	}
+	for _, rule := range profileRules {
+		ruleIdRef, err := getDsElementAttrValue(rule, "idref")
+		if err != nil {
+			return parsedProfile, fmt.Errorf("error getting value of 'idref' attribute: %s", err)
+		}
+		// change occurrences of ruleSelector to ruleSelected for alignment with selected attribute
```
marcusburghardt: I believe this comment can also be removed now.

cmd/openscap-plugin/xccdf/datastream.go
```diff
@@ -160,6 +168,36 @@ func populateProfileVariables(dsProfile *xmlquery.Node, parsedProfile *xccdf.Pro
 	return parsedProfile, nil
 }
 
+func populateProfileRules(dsProfile *xmlquery.Node, parsedProfile *xccdf.ProfileElement) (*xccdf.ProfileElement, error) {
+	if parsedProfile.Selections == nil {
+		parsedProfile.Selections = []xccdf.SelectElement{}
+	}
+	profileRules, err := getDsElements(dsProfile, "xccdf-1.2:select")
```
marcusburghardt, line 169:
```suggestion

	profileRules, err := getDsElements(dsProfile, "xccdf-1.2:select")
```
Nit, but since it is very similar to `populateProfileVariables` it would be nice to keep the empty lines consistent.

cmd/openscap-plugin/xccdf/datastream.go
```diff
@@ -170,6 +208,12 @@ func initProfile(dsProfile *xmlquery.Node, dsProfileId string) (*xccdf.ProfileEl
 	}
 
 	// Here we can add the logic to populate profile rules in a separate PR
+	// TODO: Implement the profile rules logic - to extract the Selected
```
marcusburghardt: Please, could you remove these comments since the logic was already implemented?

cmd/openscap-plugin/xccdf/datastream.go
```diff
@@ -296,7 +340,41 @@ func ResolveDsVariableOptions(profile *xccdf.ProfileElement, variables []DsVaria
 
 // Getting rule information
 // Copied from https://github.com/ComplianceAsCode/compliance-operator/blob/fed54b4b761374578016d79d97bcb7636bf9d920/pkg/utils/parse_arf_result.go#L170
+func GetDsProfileRules(dsPath string) ([]xccdf.SelectElement, error) {
```
marcusburghardt: It is nice that you worked on this rule, but it is not used in this PR, so it would be better to introduce it in a separate PR.

marcusburghardt (changes requested):
The code to populate the rules is good, but the extra function to collect rules can be part of another PR while this one focus on collecting DS Profile rules.

marcusburghardt (approved):
Thanks for working on this @hbraswelrh . All green!

cmd/openscap-plugin/xccdf/datastream.go
```diff
@@ -160,6 +161,38 @@ func populateProfileVariables(dsProfile *xmlquery.Node, parsedProfile *xccdf.Pro
...
 }
 
+func populateProfileRules(dsProfile *xmlquery.Node, parsedProfile *xccdf.ProfileElement) (*xccdf.ProfileElement, error) {
+	if parsedProfile.Selections == nil {
+		parsedProfile.Selections = []xccdf.SelectElement{}
+	}
+
+	profileRules, err := getDsElements(dsProfile, "xccdf-1.2:select")
+	if err != nil {
+		return parsedProfile, fmt.Errorf("error finding 'select' elements in profile: %s", err)
```
jpower432, line 171:
```suggestion
		return parsedProfile, fmt.Errorf("error finding 'select' elements in profile: %w", err)
```
For existing errors, you can use %w instead of %s to preserve all of the error information.

cmd/openscap-plugin/xccdf/datastream.go
```diff
@@ -160,6 +161,38 @@ func populateProfileVariables(dsProfile *xmlquery.Node, parsedProfile *xccdf.Pro
...
+	}
+
+	for _, rule := range profileRules {
+		ruleIdRef, err := getDsElementAttrValue(rule, "idref")
+		if err != nil {
+			return parsedProfile, fmt.Errorf("error getting value of 'idref' attribute: %s", err)
```
jpower432, line 177: Same here

jpower432 (changes requested):
Looks good. The only change I would recommend is to change all of the `%s` to `%w` in your returned error messages since you are adding context. This will preserve the original error object which could be useful if the error needs to be inspected by the caller.

jpower432 (approved):
LGTM. Changes will be made on the next PR.

marcusburghardt:
> LGTM. Changes will be made on the next PR.

Feedback incorporated in https://github.com/complytime/complytime/pull/37/commits/b7b2ed2c92fc5c2604072ba0a18289e3a7ac465c

PR #37, opened 2025-01-29
CPLYTM-475: Validate OSCAL rules and parameters used for openscap tailoring files

marcusburghardt:
## Summary

This PR complements #29, #33 and #35 and completes the code to generate Tailoring files to be used with `openscap-plugin`.

`oscap` consumes accurate, tested and signed Datastreams provided by `scap-security-guide` package.
These Datastream contain well tested profiles aligned to official Benchmarks and maintained by [ComplianceAsCode/content](https://github.com/complianceAsCode/content).

The Datastream should not be modified but in some cases it needs to be tailored to include or remove an specific rule or to change a variable value. In the context of OSCAL, it may also include slightly different settings when compared to a Datastream. With this PR, all rules and parameters informed by OSCAL Policy will be validated and a minimalist Tailoring files will be generated to ensure a scan aligned to OSCAL content and and the Datastream.

## Related Issues
- Relates to CPLYTM-475

## Review Hints

The commit descriptions should provide the context and more details about each change.
The features were gradually introduced by each commit, so a chronological review commit by commit tends to be easier.
The first 4 commits are more substantial while the last 2 are minor improvements and fixes identified by tests.

For unit tests
`make test-unit`

### DIFF
--- a/cmd/openscap-plugin/xccdf/tailoring.go
+++ b/cmd/openscap-plugin/xccdf/tailoring.go
@@ -135,6 +135,9 @@ func getTailoringSelections(oscalPolicy policy.Policy, dsProfile *xccdf.ProfileE
 
 func updateTailoringValues(tailoringValues, dsProfileValues []xccdf.SetValueElement, oscalPolicy policy.Policy) []xccdf.SetValueElement {
 	for _, rule := range oscalPolicy {
+		if rule.Rule.Parameter == nil {
+			continue
+		}
 		varAlreadyInDsProfile := false
 		for _, dsVar := range dsProfileValues {
 			varID := removePrefix(dsVar.IDRef, varIDPrefix)
@@ -163,6 +166,9 @@ func getTailoringValues(oscalPolicy policy.Policy, dsProfile *xccdf.ProfileEleme
 
 	// All OSCAL policy variables should be present in the Datastream
 	for _, rule := range oscalPolicy {
+		if rule.Rule.Parameter == nil {
+			continue
+		}
 		if !validateVariableExistence(rule.Rule.Parameter.ID, dsVariables) {
 			return nil, fmt.Errorf("variable not found in datastream: %s", rule.Rule.Parameter.ID)
 		}
--- a/cmd/openscap-plugin/xccdf/tailoring.go
+++ b/cmd/openscap-plugin/xccdf/tailoring.go
@@ -95,14 +95,13 @@ func selectAdditionalRules(tailoringSelections, dsProfileSelections []xccdf.Sele
 		for _, dsRule := range dsProfileSelections {
 			ruleID := removePrefix(dsRule.IDRef, ruleIDPrefix)
 			if rule.Rule.ID == ruleID {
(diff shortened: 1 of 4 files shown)


hbraswelrh (dismissed):
LGTM, great work!

cmd/openscap-plugin/xccdf/tailoring.go
```diff
@@ -49,59 +49,181 @@ func getTailoringBenchmarkHref(datastreamPath string) xccdf.BenchmarkElement {
...
 
-			values = append(values, xccdf.SetValueElement{
-				IDRef: rule.Rule.Parameter.ID,
+func selectAdditionalRules(tailoringSelections, dsProfileSelections []xccdf.SelectElement, oscalPolicy policy.Policy) []xccdf.SelectElement {
+	for _, rule := range oscalPolicy {
+		ruleAlreadyInDsProfile := false
+		for _, dsRule := range dsProfileSelections {
+			ruleID := removePrefix(dsRule.IDRef, ruleIDPrefix)
+			if rule.Rule.ID == ruleID {
+				// Not a common case, but a rule be be unselected in a Datastream Profile
```
jpower432:
```suggestion
				// Not a common case, but a rule can be unselected in a Datastream Profile
```

cmd/openscap-plugin/xccdf/tailoring.go
```diff
@@ -49,59 +49,181 @@ func getTailoringBenchmarkHref(datastreamPath string) xccdf.BenchmarkElement {
...
+
+	return tailoringSelections, nil
+}
+
+func updateTailoringValues(tailoringValues, dsProfileValues []xccdf.SetValueElement, oscalPolicy policy.Policy) []xccdf.SetValueElement {
+	for _, rule := range oscalPolicy {
+		varAlreadyInDsProfile := false
+		for _, dsVar := range dsProfileValues {
+			varID := removePrefix(dsVar.IDRef, varIDPrefix)
+			if rule.Rule.Parameter.ID == varID {
```
jpower432, line 144: Since the `Parameter` field in the `Rule` is a pointer, a panic could happen here if that value is nil since it needs to be deferenced to get the ID. Adding a check before entering the second for loop to skip any rules that don't have paramters should fix that.

jpower432 (changes requested):
Great work! Just suggesting one change and another comment with a typo fix.

cmd/openscap-plugin/xccdf/tailoring.go
marcusburghardt: Good catch @jpower432 . Let me fix this.

marcusburghardt: Fixed with https://github.com/complytime/complytime/pull/37/commits/866e00b2ebb4daa18282c5a5cbd06c46f8fdeaf3

jpower432 (approved):
LGTM.

gvauter (approved):
LGTM

PR #15, opened 2024-12-18
CPLYTM-417: add generate command

gvauter:
Adds `generate` command to CLI.

PR #38, opened 2025-01-30
fix: do not remove policy file name from config

marcusburghardt:
## Summary

With the introduction of the generate command (#15), the policy name is expected so the command can properly create the file if it is not already present.
This commit preserves the file name defined in configuration.

## Related Issues

- Avoid issues when generating a policy.

## Review Hints

This line was setting the policy file name to empty if the file was not found:
https://github.com/complytime/complytime/blob/f12079f9808680b290fd4ec0684b0c24b432c416/cmd/openscap-plugin/config/config.go#L147

However, this file is expected to be created by the `generate` command, so it can't be set to empty anymore.


jpower432 (approved):
LGTM. I tested and this solved the issue. Great work!

gvauter (approved):
LGTM

PR #46, opened 2025-02-07
CPLYTM-517: feat: add profile title to complytime list table

jpower432:
## Summary

Adds profile title to table

The Framework struct is created in the complytime pkg to hold framework data and keep interaction with the complytime app directory in the complytime pkg. The ShowDefinitionTable is now scoped to only displaying this information.

## Related Issues

CPLYTM-517

## Review Hints

**IMPORTANT** - The quick start steps have been updated the new samples should be copied before testing

To test:
```bash
make build
./bin/complytime list
```


docs/QUICK_START.md
```diff
@@ -24,6 +24,7 @@ policy validation plugin. See `docs/samples/` for example configuration for the
 
 ```bash
 cp docs/samples/sample-component-definition.json ~/.config/complytime/bundles
+cp docs/samples/sample-profile.json ~/.config/complytime/controls
```
qduanmu, line 28: Is the 'controls' directory missing under 'complytime' in L17? 'controls' dir is also created after running `complytime list`.

qduanmu (dismissed):

internal/terminal/table.go
```diff
@@ -122,5 +70,6 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 }
 
 func (m model) View() string {
-	return baseStyle.Render(m.table.View()) + "\n"
+	return helpMsg + "\n\n" + baseStyle.Render(
```
marcusburghardt: I know it is very personal and have no strong opinion on that, but it seems more intuitive to me to first show the table, which is what I would expect to see with the `list` command and then the `helpMsg`, that would be nice in case I would like to move forward with `plan`.

internal/terminal/table.go
```diff
@@ -1,36 +1,35 @@
...
-func ShowDefinitionTable(componentDefinitions []oscalTypes.ComponentDefinition) (tea.Model, error) {
-	if len(componentDefinitions) == 0 {
-		return nil, errors.New("component definitions inputs cannot be empty")
-	}
-
+// renders a table with given Framework data.
+func ShowDefinitionTable(frameworks []complytime.Framework) tea.Model {
 	columns := []table.Column{
-		{Title: "Framework ID", Width: 30},
+		{Title: "Title", Width: 30},
```
marcusburghardt, line 18: I tried to play with this creating a titles and a list of components that would be longer than the defined `Width` and the library managed it properly. I only wonder if would be valid to include a test scenario for a longer title, to have some idea on how it would be managed and shown.

internal/complytime/configuration.go
```diff
@@ -23,6 +23,7 @@ const (
 	ApplicationDir = "complytime"
 	PluginDir      = "plugins"
 	BundlesDir     = "bundles"
+	ControlsDir    = "controls"
```
marcusburghardt, line 26: Naive question, is the plan to create sub-folders for profiles and catalogs within this folder?

marcusburghardt (changes requested):
Nice work. I have some comments about user experience that don't block the PR, for your consideration.

Also, I believe the last commit can be removed in favor of rebasing the PR.

internal/complytime/configuration.go
jpower432: I think we could either do what you are suggesting or directly place both types in the `controls` folder since they will likely be accessed through the `href` of other models. In this implementation, I place a profile under `controls`.

internal/terminal/table.go
jpower432: Thanks for this feedback @marcusburghardt. I was going back and forth on this myself. I can make this change.

docs/QUICK_START.md
jpower432: Thanks @qduanmu! Good catch. I will add it.

internal/terminal/table.go
jpower432: Good idea about the test. Then if we later we want to change it, we will at least have a test to capture the current state for this use case.

marcusburghardt (approved):
LGTM. Thanks @jpower432

qduanmu (approved):
LGTM

PR #47, opened 2025-02-12
feat: CPLYTM-515 charm logger to hclog adapter

hbraswelrh:
## Summary

This PR addresses the need for an adapter that wraps the `go-hclog` hashicorp logger with the `charmbracelet/log` logger. The use of the `charmbracelet/log` package will be part of the TUI development. The `chambracelet/lipgloss` package provides styling for plugin execution output in ComplyTime.

## Related Issues
- N/A

## Review Hints

-  The commits will be labeled for each functional contribution for ease of review. 
-  Review commit by commit.
- The `log.go` adapter is coupled with the `log_test.go` unit tests. 
- Each method from `go-hclog` is wrapped in the `chambracelet/log` method that will produce a charm logger.


pkg/log/log.go
```diff
@@ -1,146 +1,167 @@
...
+	InfoColor  = lipgloss.AdaptiveColor{Light: "74", Dark: "86"}
+	WarnColor  = lipgloss.AdaptiveColor{Light: "214", Dark: "192"}
+	ErrorColor = lipgloss.AdaptiveColor{Light: "203", Dark: "203"}
+	FatalColor = lipgloss.AdaptiveColor{Light: "134", Dark: "134"}
+)
 
-// Adapt zap to hclog
-type ZapHclog struct {
-	logger *zap.SugaredLogger
+func defaultOptions() *charmlog.Options {
```
gvauter:
```suggestion
func DefaultOptions() *charmlog.Options {
```

pkg/log/log.go
```diff
@@ -1,146 +1,167 @@
...
-	"go.uber.org/zap/zapcore"
+	"strings"
 )
 
-// Enable changing the log level
-var atom = zap.NewAtomicLevel()
-
-// Pre-configure the logger
+// Initializing a New logger with default options writing to Standard error
 func init() {
```
gvauter: I'm not totally sure, but do we need to init a new logger here?  We create a new logger in the root CLI to be used throughout complytime.

pkg/log/log.go
gvauter: This can be used in other modules that create loggers.

pkg/log/log.go
```diff
@@ -1,146 +1,167 @@
...
-	return os.Stdout
+// The GetLevel() method returns the current level.
+func (c *CharmHclog) GetLevel() hclog.Level {
+	return charmHclogLevels[c.logger.GetLevel()]
 }
 
-func Logger() *zap.SugaredLogger {
-	return zap.S()
+// The StandardLog() method wraps the go-hclog StandardLogger().
+// The StandardLog() of CharmHclog is returned, wrapping go-hclog.
```
gvauter, line 162: Maybe duplicate comments here?

pkg/log/log_test.go
```diff
@@ -0,0 +1,164 @@
...
+			charmlogger.Print("")
+			assert.Equal(t, tc.expected, buf.String()) // Check if the expected string level == test val
+		})
+	}
+}
+
+// Setting level using the buffer, initializing level as Info.
+func TestNonExistentLevel(t *testing.T) {
+	var buf bytes.Buffer
+	// lo := (*CharmHclog).StandardLogger()
```
gvauter: Remove these?

gvauter (changes requested):
A few small comments, but overall looks good.

pkg/log/log.go
hbraswelrh: In that case, I don't think initializing a new logger here would make sense. I will take it out for the time being and if anything changes it can be included later.

gvauter (approved):
LGTM

PR #52, opened 2025-02-19
CPLYTM-550 - Clean-up openscap-plugin configuration file

marcusburghardt:
## Summary

- `Server` struct was only used in a very initial stage. Now the plugin is managed by ComplyTime.
- Plugin directory doesn't need to be changed through a configuration file.
  - If there is any valid case for this in the future, we can introduce options on demand. For now, it is removed in favor of simplicity.

## Related Issues

- CPLYTM-550

## Review Hints

No impact is expected on how the plugin and commands behave, so unit tests should be enough.

### DIFF
--- a/cmd/openscap-plugin/config/config_test.go
+++ b/cmd/openscap-plugin/config/config_test.go
@@ -181,14 +181,12 @@ func TestEnsureWorkspace(t *testing.T) {
 		{
 			cfg: Config{
 				Files: struct {
-					PluginDir  string "yaml:\"plugindir\""
 					Workspace  string "yaml:\"workspace\""
 					Datastream string "yaml:\"datastream\""
 					Results    string "yaml:\"results\""
 					ARF        string "yaml:\"arf\""
 					Policy     string "yaml:\"policy\""
 				}{
-					PluginDir: "plugins",
 					Workspace: filepath.Join(tempDir, "workspace"),
 					Policy:    "policy.yaml",
 					Results:   "results.xml",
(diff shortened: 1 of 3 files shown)


jpower432 (approved):
LGTM

cmd/openscap-plugin/config/config_test.go
```diff
@@ -275,9 +269,9 @@ func TestDefineFilesPaths(t *testing.T) {
 
 			if !tt.expectError {
 				// Check if the paths are correctly set
-				expectedPolicyPath := filepath.Join(tempDir, "workspace", "plugins", "policy", "policy.yaml")
-				expectedResultsPath := filepath.Join(tempDir, "workspace", "plugins", "results", "results.xml")
-				expectedARFPath := filepath.Join(tempDir, "workspace", "plugins", "results", "arf.xml")
+				expectedPolicyPath := filepath.Join(tempDir, "workspace", "openscap", "policy", "policy.yaml")
```
gvauter, line 272: Tiny nit pick - perhaps use the PluginDir constant instead of "openscap" here?  Just in case it were to ever change there would be less places to refactor.

gvauter (approved):
One small nit, but lgtm.

cmd/openscap-plugin/config/config_test.go
marcusburghardt: Good point @gvauter . I am preparing another PR on top of it and will already include this improvement. Thanks.

PR #55, opened 2025-02-20
CPLYTM-561 - Adapt scan to use tailoring file

marcusburghardt:
## Summary

With the introduction of `plan` and `generate` commands as well the availability of real content transformed from CaC to OSCAL, the `openscap-plugin` can now generate a realistic tailoring file to be used by the `scan` command.
This PR ensures all these commands work together providing the expected outcome.

## Related Issues

- CPLYTM-561

## Review Hints

The changes were introduced gradually through the commits where each commit describes the context of each change, so it should be easier to review commit by commit.

Unit tests were also adapted to the changes and `make test-unit` should be green.

### DIFF
--- a/cmd/openscap-plugin/oscap/oscap.go
+++ b/cmd/openscap-plugin/oscap/oscap.go
@@ -5,11 +5,10 @@ package oscap
 import (
 	"fmt"
 	"log"
-	"os"
 	"os/exec"
 )
 
-func constructScanCommand(openscapFiles map[string]string, profile string) ([]string, error) {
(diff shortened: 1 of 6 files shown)


jpower432 (dismissed):
LGTM. I understand why you removed the `Stat` check during scan for the tailoring file. This might be something to address later. Specifically error handling around missing policy in general so that `complytime` can inform the user that they skipped a step (`generate`).

gvauter (approved):
LGTM

jpower432 (approved):
LGTM

PR #61, opened 2025-02-25
CPLYTM-579: Treat potential issues when generating tailoring files

marcusburghardt:
## Summary

During the tests with ANSSI content there was duplicated information of rules and variables in Component Definition, which was then propagated to Assessment Plan and later to the tailoring file after the `generate` command. 
Ultimately the "scan" command failed because `oscap` scanner reported duplicated ids.
Since similar issues may happen by different reasons, a treatment was included in `openscap-plugin` to avoid errors during the `scan`.

Another case was treated in the first commit to avoid unnecessary information in the tailoring file in a case that an element is only included in the Datastream profile but already using "selected=false".

## Related Issues

- CPLYTM-579

## Review Hints

Besides the code testing, the outcome with real content can be tested using complytime-demos
Once the base VM and Ansible environment is created as described in https://github.com/complytime/complytime-demos/blob/main/README.md#get-your-hands-dirty

Populate the VM with ANSSI content used for tests from the `base_ansible_env` directory
```bash
ansible-playbook populate_complytime_anssi_content.yml
```
Connect via SSH to the Demo VM using the `ansible` user.
```bash
complytime list
complytime plan anssi_bp28_minimal
complytime generate
vim .config/complytime/openscap/policy/tailoring_policy.xml
```

### DIFF
--- a/cmd/openscap-plugin/xccdf/tailoring.go
+++ b/cmd/openscap-plugin/xccdf/tailoring.go
@@ -85,7 +85,7 @@ func unselectAbsentRules(tailoringSelections, dsProfileSelections []xccdf.Select
 				break
 			}
 		}
-		if !dsRuleAlsoInPolicy {
+		if !dsRuleAlsoInPolicy && dsRule.Selected {
 			tailoringSelections = append(tailoringSelections, xccdf.SelectElement{
 				IDRef:    dsRule.IDRef,
 				Selected: false,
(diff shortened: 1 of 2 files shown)


marcusburghardt:
It will be rebased after https://github.com/complytime/complytime/pull/55

marcusburghardt:
Rebased and ready to review.

jpower432 (approved):
LGTM. I was able to test locally using the cis profile as this fixed the duplicate id error.

gvauter (approved):
LGTM

PR #62, opened 2025-02-25
CPLYTM-512 propogate charm logger to plugins

gvauter:
## Summary
This PR adds the charm logger to the c2p config to propagate it down to the plugins.


hbraswelrh (approved):
LGTM.

PR #68, opened 2025-03-03
[CPLYTM-582 CPLYTM-583]feat: generate the assessment results as md format

huiwangredhat:
## Summary
Generate md format of assessment results 

## Related Issues
[CPLYTM-582]

## Review Hints
1. Pull latest and `make build`
2. Run` bin/complytime list`  
3. Copy the catalog.json file which was generated via trestle-bot to the ~/.config/complytime/controls/
4. Run `bin/complytime plan,  bin/complytime generate`
5. Run `bin/complytime scan `, it will generate the assement-results.json, run `bin/complytime scan -m`, it will generate assement-results.json and assement-results.md


cmd/complytime/cli/scan.go
```diff
@@ -119,14 +127,35 @@ func runScan(cmd *cobra.Command, opts *scanOptions) error {
 		return err
 	}
 
-	filePath := filepath.Join(opts.complyTimeOpts.UserWorkspace, assessmentResultsLocation)
-	cleanedPath := filepath.Clean(filePath)
-
-	err = complytime.WriteAssessmentResults(&assessmentResults, cleanedPath)
-	if err != nil {
-		return err
+	switch outputFormat {
```
gvauter: @huiwangredhat for the output I believe we will always want to write the JSON file as this is required for remediation.  The markdown file could be me optional, or just always written as well (at least for now).

gvauter (changes requested):
The assessment-results.json file will always need to be written.  The human readable markdown could either also be written on every scan or perhaps a flag to enable/disable?

internal/complytime/configuration.go
```diff
@@ -64,6 +67,7 @@ func newApplicationDirectory(rootDir string, create bool) (ApplicationDirectory,
 	applicationDir.pluginDir = filepath.Join(applicationDir.appDir, PluginDir)
 	applicationDir.bundleDir = filepath.Join(applicationDir.appDir, BundlesDir)
 	applicationDir.controlDir = filepath.Join(applicationDir.appDir, ControlsDir)
+	applicationDir.catalogDir = filepath.Join(applicationDir.appDir, CatalogsDir)
```
jpower432: Nit: Could the `catalogDir` be a sub directory of `controlDir` since catalogs are in the OSCAL controls layer?

cmd/complytime/cli/scan.go
jpower432: This is a good point @gvauter. However, I do think it is worth it to keep markdown optional for the automation/scripting use case. Maybe the `--output` flag would just become a boolean to conditionally write the report?

jpower432 (commented):
Looks good @huiwangredhat. I added to the discussion @gvauter started and had one small comment about the app directory structure.

cmd/complytime/cli/scan.go
huiwangredhat: Thanks for all your feedback. I'd like to update it to --output with boolean, if it is false, it won't write the md, else it will write the md. But anyway it will write the json.

huiwangredhat: Run `complytime scan -o`, it will generate the md, else it won't generate the md. Both will generate json.

internal/complytime/configuration.go
huiwangredhat: I removed the catalogDir, and located the catalog.json to the controls dir.

internal/complytime/catalogs.go
```diff
@@ -0,0 +1,32 @@
...
+	oscalTypes "github.com/defenseunicorns/go-oscal/src/types/oscal-1-1-2"
+	"github.com/oscal-compass/oscal-sdk-go/generators"
+)
+
+// The assessment results md file needs the catalog title information
+// LoadCatalogSource returns an OSCAL catalogs from a given application directory.
+
+func LoadCatalogSource(appDir ApplicationDirectory) (*oscalTypes.Catalog, error) {
+	var path string
+	path = "controls/catalog.json"
```
jpower432: I missed this on my first review - I think there will be more than one catalog in the `controls` directory in most use cases.

I think to get the correct catalog you would need to load it through the profile [imports](https://pages.nist.gov/OSCAL-Reference/models/v1.1.3/profile/json-reference/#/profile/imports).

To the profile href might be a bit trickier, but one approach to get it is the `implementation.Source` on the Control Implementation [here](https://github.com/complytime/complytime/blob/ba8293b63c545dde01b965e65be5fef5410fc3b0/cmd/complytime/cli/scan.go#L106). What do you think?

cmd/complytime/cli/scan.go
jpower432: I tested and this worked as described!

internal/complytime/catalogs.go
huiwangredhat: Thank you! I have tried to merge the example commit to this PR but failed. So I updated it according to your example and tested it. At that stage, the catalog is loaded by the imports[0]. Later it needs to be improved if more components/profiles are supported.

internal/complytime/testdata/complytime/controls/sample-catalog.json
```diff
@@ -0,0 +1,42 @@
...
+                "id": "r1_smt",
+                "name": "statement"
+              }
+            ]
+          }
+        ]
+      }
+    ]
+  }
+}
```
marcusburghardt: Nit: Could you include an empty line at the end, please?

cmd/complytime/cli/scan.go
```diff
@@ -42,6 +44,7 @@ func scanCmd(common *option.Common) *cobra.Command {
 			return runScan(cmd, scanOpts)
 		},
 	}
+	cmd.Flags().BoolP("output", "o", false, "If true, assessement-result markdown will be generated")
```
marcusburghardt: Maybe we should rename the flag to more intuitively say that the output would be a Markdown instead of only "output". Markdown the only optional output for now and no other formats are expected in the short to medium term, so this minor change could make the experience better.

For example:
```complytime scan --with-md```

What do you think?

**Review by marcusburghardt — COMMENTED** [review]
I have one non-blocking comment that would like to hear your thoughts.

File: cmd/complytime/cli/scan.go (hunk 5 above)
Comment (huiwangredhat) [reply cmd/complytime/cli/scan.go]: That's reasonable. I updated it.

**Review by marcusburghardt — DISMISSED** [review]
LGTM. Thanks for the updates @huiwangredhat

File: cmd/complytime/cli/scan.go [hunk 6]
```diff
@@ -98,13 +101,26 @@ func runScan(cmd *cobra.Command, opts *scanOptions) error {
...
 	for _, compDef := range cfg.ComponentDefinitions {
 		for _, component := range *compDef.Components {
 			if component.ControlImplementations == nil {
 				continue
 			}
-			allImplementations = append(allImplementations, *component.ControlImplementations...)
-
+			for _, implementation := range *component.ControlImplementations {
+				frameworkShortName, found := settings.GetFrameworkShortName(implementation)
+				fmt.Println(frameworkShortName)
```
jpower432: Nit: Are the included print statements intentional?

jpower432 (dismissed):
LGTM. One tiny nit/question about the inclusion of some of the print statements.

cmd/complytime/cli/scan.go
gvauter: Just to add to this, if desired, the `logger` work has now been merged and available here if you'd like to debug log.

marcusburghardt:
@huiwangredhat , could you remove the last commit (merge) and rebase the PR with the latest main branch, please?
```
git checkout main
git pull upstream main # Replace upstream by the name you use targeting https://github.com/complytime/complytime
git checkout generate_md_from_scan
git rebase main
```
Something like this should work. : )

cmd/complytime/cli/scan.go
```diff
@@ -125,14 +139,44 @@ func runScan(cmd *cobra.Command, opts *scanOptions) error {
...
+		if err != nil {
+			return err
+		}
+		err = os.WriteFile(cleanedPath, assessmentResultsMd, 0600)
+		if err != nil {
+			return err
+		}
+		logger.Info(fmt.Sprintf("The assessment results were successfully written to %v.", cleanedPath))
+	} else {
+		fmt.Println("No assessment result markdown will be generated.")
```
marcusburghardt: While testing it locally, this message seems not necessary. Maybe it could be informed only in debug (logger.Debug) or remove the `else` completely. What do you think?

cmd/complytime/cli/scan.go
```diff
@@ -125,14 +139,44 @@ func runScan(cmd *cobra.Command, opts *scanOptions) error {
 	if err != nil {
 		return err
 	}
-
-	filePath := filepath.Join(opts.complyTimeOpts.UserWorkspace, assessmentResultsLocation)
+	filePath := filepath.Join(opts.complyTimeOpts.UserWorkspace, assessmentResultsLocationJson)
```
marcusburghardt: `filePath` is used here and below with a different assignment. For readability, it could be good to rename it for something like `arJsonPath` here.

cmd/complytime/cli/scan.go
```diff
@@ -125,14 +139,44 @@ func runScan(cmd *cobra.Command, opts *scanOptions) error {
...
+		}
+
+		if len(profile.Imports) != 1 {
+			return errors.New("profile imports must be one")
+		}
+		catalog, err := complytime.LoadCatalogSource(appDir, profile.Imports[0].Href)
+		if err != nil {
+			return err
+		}
+		filePath := filepath.Join(opts.complyTimeOpts.UserWorkspace, assessmentResultsLocationMd)
```
marcusburghardt: `filePath` is used here and above with a different assignment. For readability, it could be good to rename it for something like `arMarkdownPath` here.

cmd/complytime/cli/scan.go
```diff
@@ -125,14 +139,44 @@ func runScan(cmd *cobra.Command, opts *scanOptions) error {
...
+		cleanedPath := filepath.Clean(filePath)
```
marcusburghardt: At this point `opts.complyTimeOpts.UserWorkspace` is already cleaned and `assessmentResultsLocationMd` is a const. Therefore `cleanedPath := filepath.Clean(filePath)` is not necessary. The same is applicable for the Json file above. Knowing it was already there before, I think is a good opportunity to simplify.

internal/complytime/catalogs.go
```diff
@@ -0,0 +1,20 @@
...
+package complytime
+
+import (
+	oscalTypes "github.com/defenseunicorns/go-oscal/src/types/oscal-1-1-2"
+	"github.com/oscal-compass/oscal-sdk-go/generators"
+)
+
+// The assessment results md file needs the catalog title information
+// LoadCatalogSource returns an OSCAL catalogs from a given application directory.
+
```
marcusburghardt:
```suggestion
```
Nit: this extra empty line could be removed to be consistent with other comments.

internal/complytime/controls.go
```diff
@@ -94,8 +95,18 @@ func processComponent(appDir ApplicationDirectory, component oscalTypes.DefinedC
 	return frameworks, nil
 }
 
+// LoadProfileSource returns an OSCAL profile from a given control source from a given application directory.
```
marcusburghardt:
```suggestion
// LoadProfile returns an OSCAL profile from a given control source from a given application directory.
```

internal/complytime/controls.go
```diff
@@ -94,8 +95,18 @@ func processComponent(appDir ApplicationDirectory, component oscalTypes.DefinedC
...
+	if err != nil {
+		return nil, err
+	}
+	defer sourceFile.Close()
+	return generators.NewProfile(sourceFile)
+}
+
 // LoadControlSource returns an OSCAL profiles from a given control source from a given application directory.
-func LoadControlSource(appDir ApplicationDirectory, controlSource string) (*oscalTypes.Profile, error) {
+func findControlSource(appDir ApplicationDirectory, controlSource string) (io.ReadCloser, error) {
```
marcusburghardt, line 109: After changing the function, the comment doesn't seem correct. Could you update it, please?

internal/complytime/testdata/complytime/controls/sample-profile.json
```diff
@@ -9,7 +9,7 @@
     },
     "imports": [
       {
-        "href": "file://catalog.json",
+        "href": "file://contorls/sample-catalog.json",
```
marcusburghardt:
```suggestion
        "href": "file://controls/sample-catalog.json",
```

marcusburghardt (changes requested):
Thanks @huiwangredhat . The code is working on my local tests. But besides some minor issues, there are also some opportunities for simplification. Please, take a look in the comments.

huiwangredhat:
@marcusburghardt Thanks for the suggestions. I updated the code according to your comments.

marcusburghardt (commented):
Thanks for the updates @huiwangredhat . LGTM, but we need to rebase before we can merge it. Could you take a look, please?

jpower432 (approved):
LGTM

marcusburghardt (approved):
Thanks for working on this PR and promptly addressing the conflicts @huiwangredhat . Nice work.

gvauter (approved):
LGTM

PR #73, opened 2025-03-07
Include pre-commit hook for SPDX line

marcusburghardt:
## Summary

This is an alternative in Python for the https://github.com/complytime/complytime/pull/18.
I had started some tests with this a couple of months ago and would like to share it so we can consider the option as well.

Some additional differences is that this script will receive the relevant files directly from pre-commit, which is declared already with filters to only consider relevant files based on their extensions. This should be more flexible to extend and is not necessary to use external commands to filter files, such as `git ls-files`.
In Python we can also better treat some possible errors.

## Related Issues

- There is no issue related to this PR, but is an improvement for contributors.

## Review Hints

You can create a source file without SPDX line and try to commit it. For example:
```
cat << EOF > scripts/test.sh
#!/bin/bash

echo "test"

EOF

git add scripts/test.sh
git commit
```
Another alternative would be to test the script directly:
```
python scripts/check-and-fix-spdx.py scripts/test.sh 
Fixed SPDX line in scripts/test.sh
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,3 +18,12 @@ repos:
       #- id: no-go-testing
       #- id: golangci-lint
       - id: go-unit-tests
+
+-   repo: local
+    hooks:
+        - id: check-spdx
+          name: Check and fix SPDX line in source files
+          entry: python ./scripts/check-and-fix-spdx.py
+          language: system
+          stages: [pre-commit]
+          files: \.(py|go|sh)$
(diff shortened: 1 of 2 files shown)


beatrizmcouto:
@d10n started review, to test. One more review missing @RichardXuan @qduanmu @huiwangredhat does one of you want to volunteer for the other review??

huiwangredhat (approved):
LGTM.  I tested sh, go, and py,  all of them work well.

d10n (approved):

d10n:
I'd prefer if the hook is pre-push, so that there is less friction in making local WIP commits, but this is good

hbraswelrh (approved):
LGTM

PR #80, opened 2025-04-03
CPLYTM-591 feat: use hclog in openscap-plugin CPLYTM-696 CPLYTM-697

d10n:
## Summary
Use hclog in openscap-plugin

## Review Hints
I decided to use a global logger for simplicity, but wouldn't be opposed to making a log package in the plugin

### DIFF
--- a/cmd/openscap-plugin/main.go
+++ b/cmd/openscap-plugin/main.go
@@ -3,13 +3,26 @@
 package main
 
 import (
+	"github.com/hashicorp/go-hclog"
+
 	"github.com/complytime/complytime/cmd/openscap-plugin/server"
 
 	hplugin "github.com/hashicorp/go-plugin"
 	"github.com/oscal-compass/compliance-to-policy-go/v2/plugin"
 )
 
+var log hclog.Logger
+
+func init() {
+	log = hclog.New(&hclog.LoggerOptions{
+		Name:  "openscap-plugin",
+		Level: hclog.Debug,
+	})
+	hclog.SetDefault(log)
+}
+
 func main() {
+	hclog.Default().Info("Starting OpenSCAP plugin")
 	openSCAPPlugin := server.New()
 	pluginByType := map[string]hplugin.Plugin{
 		plugin.PVPPluginName: &plugin.PVPPlugin{Impl: openSCAPPlugin},
--- a/cmd/openscap-plugin/main.go
+++ b/cmd/openscap-plugin/main.go
@@ -11,14 +11,14 @@ import (
 	"github.com/oscal-compass/compliance-to-policy-go/v2/plugin"
 )
 
-var log hclog.Logger
+var logger hclog.Logger
 
 func init() {
-	log = hclog.New(&hclog.LoggerOptions{
+	logger = hclog.New(&hclog.LoggerOptions{
 		Name:  "openscap-plugin",
 		Level: hclog.Debug,
 	})
-	hclog.SetDefault(log)
+	hclog.SetDefault(logger)
 }
 
 func main() {
--- a/cmd/openscap-plugin/main.go
+++ b/cmd/openscap-plugin/main.go
@@ -3,6 +3,8 @@
 package main
 
 import (
+	"os"
+
 	"github.com/hashicorp/go-hclog"
 
 	"github.com/complytime/complytime/cmd/openscap-plugin/server"
@@ -15,8 +17,10 @@ var logger hclog.Logger
 
 func init() {
 	logger = hclog.New(&hclog.LoggerOptions{
-		Name:  "openscap-plugin",
-		Level: hclog.Debug,
+		Name:       "openscap-plugin",
+		Level:      hclog.Debug,
+		Output:     os.Stderr,
+		JSONFormat: true,
 	})
 	hclog.SetDefault(logger)
 }
--- a/cmd/openscap-plugin/oscap/oscap.go
+++ b/cmd/openscap-plugin/oscap/oscap.go
@@ -4,8 +4,9 @@ package oscap
 
 import (
 	"fmt"
-	"log"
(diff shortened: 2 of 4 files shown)


cmd/openscap-plugin/main.go
```diff
@@ -3,13 +3,26 @@
...
 import (
+	"github.com/hashicorp/go-hclog"
+
 	"github.com/complytime/complytime/cmd/openscap-plugin/server"
 
 	hplugin "github.com/hashicorp/go-plugin"
 	"github.com/oscal-compass/compliance-to-policy-go/v2/plugin"
 )
 
+var log hclog.Logger
```
marcusburghardt: In complytime cli it is used `logger`. Maybe we could also use `logger` here instead of `log` to be consistent.
https://github.com/complytime/complytime/blob/main/cmd/complytime/cli/root.go#L15

cmd/openscap-plugin/main.go
```diff
@@ -3,13 +3,26 @@
...
+func init() {
+	log = hclog.New(&hclog.LoggerOptions{
+		Name:  "openscap-plugin",
+		Level: hclog.Debug,
+	})
+	hclog.SetDefault(log)
+}
+
 func main() {
+	hclog.Default().Info("Starting OpenSCAP plugin")
```
marcusburghardt, line 29: I think this could be simplified to `log.Info("Starting OpenSCAP plugin")`.

cmd/openscap-plugin/server/server.go
```diff
@@ -46,7 +47,7 @@ func (s PluginServer) Configure(configMap map[string]string) error {
 }
 
 func (s PluginServer) Generate(policy policy.Policy) error {
-	fmt.Println("Generating a tailoring file")
+	hclog.Default().Info("Generating a tailoring file")
```
marcusburghardt, line 50: The messages from plugin are only visible in debug mode.
```
INFO openscap-plugin: configuring server automatic mTLS timestamp=2025-04-07T14:46:11.067Z
DEBUG openscap-plugin: plugin address address=/tmp/plugin2622711003 network=unix timestamp=2025-04-07T14:46:11.076Z
DEBUG using plugin version=1
DEBUG Launched plugin openscap
DEBUG Gathering configuration options for openscap
DEBUG stdio: waiting for stdio data
DEBUG Generating policy for provider openscap
DEBUG openscap-plugin: 2025-04-07T14:46:11.395Z [INFO]  openscap-plugin: Generating a tailoring file
INFO Policy generation process completed for available plugins.
```
It would be more intuitive if we could call something like [logger.info()](https://github.com/complytime/complytime/blob/main/cmd/complytime/cli/generate.go#L91) or [logger.debug()](https://github.com/complytime/complytime/blob/main/cmd/complytime/cli/generate.go#L65) to defined when to show each message.

What do you think?

marcusburghardt (commented):
@d10n , I left some initial comments to discuss/consider. What do you think?

cmd/openscap-plugin/main.go
d10n: (as discussed) I chose to get the global logger like this so that it is greppable along with the other openscap-plugin logs which don't have the log variable available

cmd/openscap-plugin/server/server.go
d10n: I think this is resolved now

cmd/openscap-plugin/oscap/oscap.go
```diff
@@ -36,18 +37,18 @@ func OscapScan(openscapFiles map[string]string, profile string) ([]byte, error)
...
 	if err != nil {
 		if err.Error() == "exit status 1" {
 			return output, fmt.Errorf("%s: oscap error during evaluation", err)
 		} else if err.Error() == "exit status 2" {
-			log.Printf("%s: at least one rule resulted in fail or unknown", err)
+			hclog.Default().Error("at least one rule resulted in fail or unknown", "err", err)
 			return output, nil
 		} else {
-			log.Printf("%s", err)
+			hclog.Default().Error("Error", "err", err)
```
jpower432: Nit: Seems like an error should be returned here instead of logged.

cmd/openscap-plugin/oscap/oscap.go
```diff
@@ -36,18 +37,18 @@ func OscapScan(openscapFiles map[string]string, profile string) ([]byte, error)
...
+	hclog.Default().Info("Executing command", "command", command)
 	cmd := exec.Command(cmdPath, command[1:]...)
 
 	output, err := cmd.CombinedOutput()
 	if err != nil {
 		if err.Error() == "exit status 1" {
 			return output, fmt.Errorf("%s: oscap error during evaluation", err)
 		} else if err.Error() == "exit status 2" {
-			log.Printf("%s: at least one rule resulted in fail or unknown", err)
+			hclog.Default().Error("at least one rule resulted in fail or unknown", "err", err)
```
jpower432: Nit: Should this be an an error or a warning? @marcusburghardt Did you have thoughts on the level here?

jpower432 (dismissed):
LGTM. Two non-blocking nits that can be addressed in a follow-on if needed.

cmd/openscap-plugin/oscap/oscap.go
d10n: I think these are business logic decisions out of scope but I agree it makes sense to review in a follow-up

cmd/openscap-plugin/oscap/oscap.go
marcusburghardt: A warning is more appropriated here. It is expected that a not yet hardened system results in fail for at least for one rule.

jpower432 (approved):
LGTM

marcusburghardt (approved):
LGTM

PR #84, opened 2025-04-10
feat: CPLYTM-659 introducing the config option for assessment plan editing (WIP)

hbraswelrh:
## Summary

This PR introduces the `update.go` file to populate a `config.yaml` in the complytime workspace. The changes made will allow the complytime user to copy and paste the default configuration from StdOut to the command line. Once pasting back to the command line - the default fields can be updated to tailor the assessment plan as preferred. The changes will then be leveraged to update the assessment-plan.  

The default config values are printed to standard out as `key: value` pairs.  The inclusion of both the `config.json` and `config.yaml` are for example purposes. 

## Related Issues

- CPLYTM-659

## Review Hints

-  The data used is for the purpose of an example. The functionality will be transitioned to only `.JSON` or `.yml` for populating the config file.

- Running `go run main.go` will populate the example in the output. 
-   image
-  image
-   image

**Miro Board**
-  image
- [Screen recording link](https://drive.google.com/drive/folders/1ykqQV2YpFK_Ppk982k8B8vojCrlhwFtg?usp=drive_link)

### DIFF
--- a/cmd/complytime/cli/update.go
+++ b/cmd/complytime/cli/update.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-
 )
 
 // Structure for marshalling a go struct into json data
@@ -14,6 +13,7 @@ type Config struct {
 	IncludeControls []string `json:"include_controls"`
 	ControlIds      []string `json:"control_ids"`
 }
+
 // Function called in main that will populate config.json defaults
 func Write() {
 	c := Config{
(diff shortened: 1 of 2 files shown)


hbraswelrh:
closing due to duplication error

jpower432:
I think this option has the potential to be the one of simplest to implement and simplest to use for someone that does not prefer a UI for guided editing.

Some use cases where I see this working better over flags:

1. Building an assessment plan might be an iterative process. Due to the ephemeral nature of command line flags, this could be challenging unless the command is used in a script.
2. Configs provide support for complex data types which are uses here. Flags could as well, but it may not result in the most readable command.
3. Configs would be useful in the event a user wanted to regenerate an assessment plan exactly or create a similar assessment with slight modifications.

PR #91, opened 2025-04-14
fix: set `make all` as default target

d10n:
## Summary
`make`  executes the first target, which is conventionally something like `all` or `build`


jpower432 (approved):
LGTM

gvauter (approved):
LGTM

PR #113, opened 2025-04-30
feat: CPLYTM-724: update ApplicationDirectory for plugin manifest

qduanmu:
## Summary
This is a downstream update in complytime for separating plugin binary from manifest.

## Related Issues
- Closes CPLYTM-724


internal/complytime/configuration.go
```diff
@@ -48,7 +51,12 @@ type ApplicationDirectory struct {
 // If the application directories exist, this will not overwrite what is
 // existing.
 func NewApplicationDirectory(create bool) (ApplicationDirectory, error) {
-	return newApplicationDirectory(xdg.ConfigHome, create)
+	// When complytime installed by rpm package
+	if _, err := os.Stat("/usr/bin/complytime"); err == nil {
+		return newApplicationDirectory(DataRootDir, false)
+	} else {
+		return newApplicationDirectory(xdg.ConfigHome, create)
```
jpower432: I wonder if it would make more sense to change this to xdg.DataHome given that is it analogous to `/usr/share`. Might be out of scope for this PR though.

internal/complytime/configuration.go
```diff
@@ -48,7 +51,12 @@ type ApplicationDirectory struct {
 // If the application directories exist, this will not overwrite what is
 // existing.
 func NewApplicationDirectory(create bool) (ApplicationDirectory, error) {
-	return newApplicationDirectory(xdg.ConfigHome, create)
+	// When complytime installed by rpm package
+	if _, err := os.Stat("/usr/bin/complytime"); err == nil {
```
jpower432: This is a nice addition for backward compatibility. Depending on the use case, it may be nice to control through through an environment variable. What do you think?

I imagine we would only continue to use the original location for the application directory during development. So maybe `COMPLYTIME_DEV_MODE`? If I am developing locally, I might have a stable version of `complytime` installed with an RPM and also testing another version locally.

jpower432 (commented):
Looks good @qduanmu. Leaving some initial feedback. None of the feedback is blocking, just discussion points. I was able to verify that an existing setup from the quick start still works as expected and I need to verify the RPM install.

gvauter (dismissed):
LGTM

complytime.spec
```diff
@@ -38,24 +38,31 @@ mkdir -p %{buildroot}%{_libexecdir}/%{name}/plugins
...
 make test-unit
 
 %files
-%license LICENSE
 %{_bindir}/complytime
-%doc README.md
 %defattr(-,root,root,644)
-%{_datadir}/%{name}
+%license LICENSE
+%doc README.md
```
marcusburghardt: For docs we could already use the man pages created here: https://github.com/complytime/complytime/pull/83

jpower432 (changes requested):
Thanks for making the changes @qduanmu. I was able to run locally in dev mode and not in dev mode successfully. One pre-step I needed to do for dev mode was `cp -r  ~/.config/complytime/ ~/.local/share/`. So I would suggest we update the quick start and scripts.

marcusburghardt (commented):
@qduanmu , I included two commits on top of this PR to update the source version and some other few improvements regarding man and installed files in spec file. Tests are ok but I saw there are 3 merge commits included in the PR that ideally we should remove before merging it.

marcusburghardt (approved):
LGTM. Thanks @qduanmu

jpower432 (approved):
LGTM

PR #127, opened 2025-05-13
feat: introducing the write function to update plan CPLYTM-237 CPLYTM-758 CPLYTM-757

d10n:
## Summary
Adds `plan --dry-run` for generating the assessment plan filtering config


go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	github.com/spf13/cobra v1.8.1
 	github.com/spf13/pflag v1.0.6-0.20210604193023-d5e0c0615ace
 	github.com/stretchr/testify v1.10.0
+	gopkg.in/yaml.v3 v3.0.1
```
jpower432: This project is archived. Perhaps switch out for `github.com/goccy/go-yaml`.

cmd/complytime/cli/root.go
```diff
@@ -44,12 +44,23 @@ func New() *cobra.Command {
...
+	//	planCmd_,
+	//	listCmd(&opts),
+	//)
 
 	cmd.AddCommand(
 		versionCmd(&opts),
 		scanCmd(&opts),
 		generateCmd(&opts),
 		planCmd(&opts),
+		//updateCmd(&opts),
```
jpower432: Can this be removed?

cmd/complytime/cli/root.go
```diff
@@ -44,12 +44,23 @@ func New() *cobra.Command {
 		},
 	}
 	opts.BindFlags(cmd.PersistentFlags())
+	//
```
jpower432: Can this be removed?

cmd/complytime/cli/plan_dry_run.go
```diff
@@ -0,0 +1,56 @@
...
+
+// planDryRun leverages the PlanData structure to populate tailoring config.
+// The config is written to stdout.
+func planDryRun(frameworkId string, cds []oscalTypes.ComponentDefinition) {
+	basePlanData := PlanData{
+		FrameworkID: frameworkId,
+		Controls:    []string{},
+	}
+	if cds == nil {
+		fmt.Fprintln(os.Stderr, "no component definitions found")
```
jpower432: Curious about the approach to use a bare return instead of returning an error here. Does the plan command not use `RunE`?

cmd/complytime/cli/plan_dry_run.go
```diff
@@ -0,0 +1,56 @@
...
+						basePlanData.Controls = append(basePlanData.Controls, ir.ControlId)
+					}
+				}
+			}
+		}
+	}
+
+	out, err := yaml.Marshal(&basePlanData)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "error marshalling yaml content: ", err)
```
jpower432: Same comment about returning an error. Is the goal to print `out` whether there is an error or not?

jpower432 (changes requested):
Looks good. Added a few suggestions.

cmd/complytime/cli/plan_dry_run.go
d10n: I thought that this might not necessarily be an error, but I'm also fine treating it as an error

cmd/complytime/cli/plan_dry_run.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	for _, componentDef := range cds {
+		if componentDef.Components == nil {
+			continue
+		}
+		for _, component := range *componentDef.Components {
+			if component.ControlImplementations == nil {
+				continue
+			}
+			for _, ci := range *component.ControlImplementations {
```
jpower432, line 37: It seems that if you don't filter at the `ControlImplementation` level, controls will be added to the plan data that are not part of the chosen framework ID. Will that be address later on?

jpower432 (commented):
Thanks for making the changes. Everything looks good, just one question before I approve.

jpower432:
This logic has been incorporated into PR #134

PR #133, opened 2025-05-22
CPLYTM-748 CPLYTM-749 CPLYTM-761 CPLYTM-788 Add info command

gvauter:
## Summary
This PR adds a the `info` command to comlytime.  The purpose of this command is to allows users to inspect the controls and rules of a given framework.  

Example usage:

`complytime info <framework-id>` // shows the controls within the framework
`complytime info <framework-id>  --control <control-id>` // shows details about a given control
`complytime info <framework-id>  --rule <rule-id>` // shows details about a given control

## Review Hints

https://github.com/complytime/complytime/pull/133/commits/61fb4c23ef1ee4e23bd0ddff0dabfc89c055f3a7 contains changes to refactor the existing `table.go` module for re-use with the info command
https://github.com/complytime/complytime/pull/133/commits/d694f8b3e9623cbc16b11672a4a056a027249f44 contains the actual implementation of the `info` command
4b411687c09f2ff78ab8013fb4162a84436c0cad adds support for plain table generation

### DIFF
--- a/cmd/complytime/cli/info.go
+++ b/cmd/complytime/cli/info.go
@@ -25,7 +25,6 @@ import (
 )
 
 const (
-	defaultTableLimit      = 10
 	colWidthControlID      = 12
 	colWidthControlTitle   = 40
 	colWidthImplStatus     = 21
@@ -65,8 +64,10 @@ var (
 // Rule represents details about a rule for easy
 // mapping of rules to plugins.
 type Rule struct {
-	ID     string
-	Plugin string
+	ID          string
+	Plugin      string
+	Description string
+	Parameters  []string
 }
 
 // Control repsents details about a control across
@@ -77,15 +78,6 @@ type Control struct {
 	Description         string
 	ImplemenationStatus string
 	Rules               []Rule
-}
-
-// RuleInfo represents details about a rule
-type RuleInfo struct {
-	ID               string
-	Description      string
-	CheckID          string
-	CheckDescription string
-	Parameters       []string
 }
 
 type infoOptions struct {
@@ -116,7 +108,7 @@ func infoCmd(common *option.Common) *cobra.Command {
 	}
 	cmd.Flags().StringVarP(&infoOpts.controlID, "control", "c", "", "show info for a specific control ID")
 	cmd.Flags().StringVarP(&infoOpts.ruleID, "rule", "r", "", "show info for a specific rule ID")
-	cmd.Flags().IntVarP(&infoOpts.limit, "limit", "l", 10, fmt.Sprintf("limit the number of table rows (default is %d)", defaultTableLimit))
+	cmd.Flags().IntVarP(&infoOpts.limit, "limit", "l", 0, fmt.Sprintf("limit the number of table rows"))
 	cmd.Flags().BoolVarP(&infoOpts.plain, "plain", "p", false, "print the table with minimal formatting")
 	infoOpts.complyTimeOpts.BindFlags(cmd.Flags())
 	return cmd
@@ -196,9 +188,13 @@ func processControlImplementations(compDefs []oscalTypes.DefinedComponent, appDi
 
 					if ir.Props != nil {
 						for _, p := range *ir.Props {
-							if p.Name == "Rule_Id" && p.Value != "" {
-								controlDetails.Rules = append(controlDetails.Rules, Rule{ID: p.Value, Plugin: pluginForComponent})
-							} else if p.Name == "implementation-status" && p.Value != "" {
+							switch p.Name {
+							case "Rule_Id":
+								rule := extractRuleDetails(*ir.Props)
+								rule.ID = p.Value
+								rule.Plugin = pluginForComponent
+								controlDetails.Rules = append(controlDetails.Rules, rule)
+							case "implementation-status":
 								controlDetails.ImplemenationStatus = p.Value
 							}
 						}
@@ -303,16 +299,12 @@ func filterFrameworkComponents(componentDefinitions []oscalTypes.ComponentDefini
 }
 
 // extractRuleDetails parses properties to fill a RuleInfo struct.
-func extractRuleDetails(props []oscalTypes.Property) RuleInfo {
-	info := RuleInfo{}
+func extractRuleDetails(props []oscalTypes.Property) Rule {
+	info := Rule{}
 	for _, prop := range props {
 		switch prop.Name {
 		case "Rule_Description":
 			info.Description = prop.Value
-		case "Check_Id":
-			info.CheckID = prop.Value
-		case "Check_Description":
-			info.CheckDescription = prop.Value
 		case "Parameter_Id":
 			info.Parameters = append(info.Parameters, prop.Value)
 		}
@@ -384,7 +376,7 @@ func getControlRulesColumnsAndRows(control Control) ([]table.Column, []table.Row
 }
 
 // getRuleParametersColumnsAndRows prepares columns and rows for the rule parameters table.
-func getRuleParametersColumnsAndRows(ruleDetails RuleInfo, setParameters map[string][]string) ([]table.Column, []table.Row) {
+func getRuleParametersColumnsAndRows(ruleDetails Rule, setParameters map[string][]string) ([]table.Column, []table.Row) {
 	var rows []table.Row
 
 	// Sort parameters by ID for consistent ordering in the table
@@ -447,7 +439,7 @@ func getControlListColumnsAndRows(controls []Control) ([]table.Column, []table.R
 	columns := []table.Column{
 		{Title: "Control ID", Width: colWidthControlID},
 		{Title: "Control Title", Width: colWidthControlTitle},
-		{Title: "Implementation Status", Width: colWidthImplStatus},
+		{Title: "Status", Width: colWidthImplStatus},
 		{Title: "Plugins Used", Width: colWidthPluginsUsed},
 	}
 
@@ -474,7 +466,7 @@ func newControlInfoModel(control Control, rowLimit int) terminal.Model {
 	headerFields := strings.Join([]string{
 		renderKeyValuePair("Control ID", control.ID),
 		renderKeyValuePair("Title", control.Title),
-		renderKeyValuePair("Implementation Status", control.ImplemenationStatus),
+		renderKeyValuePair("Status", control.ImplemenationStatus),
 		keyStyle.Render("Description") + ":\n" + valueStyle.Render(wrappedDescription),
 	}, "\n")
 
@@ -482,16 +474,13 @@ func newControlInfoModel(control Control, rowLimit int) terminal.Model {
 
 	columns, rows := getControlRulesColumnsAndRows(control)
 
-	effectiveRuleLimit := rowLimit
-	if effectiveRuleLimit > len(control.Rules) {
-		effectiveRuleLimit = len(control.Rules)
-	}
+	tableHeight := calculateRowLimit(rowLimit, len(rows))
 
 	tbl := table.New(
 		table.WithColumns(columns),
 		table.WithRows(rows),
 		table.WithFocused(true),
-		table.WithHeight(effectiveRuleLimit+1), // +1 for header row
+		table.WithHeight(tableHeight),
 	)
 
 	tbl.SetStyles(table.Styles{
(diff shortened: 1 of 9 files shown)


cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,619 @@
...
+
+	"github.com/spf13/cobra"
+
+	"github.com/complytime/complytime/cmd/complytime/option"
+	"github.com/complytime/complytime/internal/complytime"
+	"github.com/complytime/complytime/internal/terminal"
+)
+
+const (
+	defaultTableLimit      = 10
```
marcusburghardt: Most of the frameworks has hundreds of controls, so I assume the user will always repeat the command using `--limit`. Maybe we could change the default to dynamically show all the controls unless the user asks to limit the output. What do you think?

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,619 @@
...
+	Title               string
+	Description         string
+	ImplemenationStatus string
+	Rules               []Rule
+}
+
+// RuleInfo represents details about a rule
+type RuleInfo struct {
+	ID               string
+	Description      string
```
marcusburghardt: Maybe we could unify the rule properties in `Rule` struct, by moving `Description` and `Parameters` there? The `CheckID` and `CheckDescription` are currently the same information from `RuleID` and `RuleDescription`. Maybe this will continue this way for some time, so we could probably remove the information from the struct.

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,619 @@
...
+							ID:          ir.ControlId,
+							Title:       controlTitle,
+							Description: ir.Description,
+							Rules:       []Rule{},
+						}
+					}
+
+					if ir.Props != nil {
+						for _, p := range *ir.Props {
+							if p.Name == "Rule_Id" && p.Value != "" {
```
marcusburghardt: Considering the `Rules` struct is unified, we could also populate the `Description` here by reading the `Rule_Description`.

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,619 @@
...
+}
+
+// extractRuleDetails parses properties to fill a RuleInfo struct.
+func extractRuleDetails(props []oscalTypes.Property) RuleInfo {
+	info := RuleInfo{}
+	for _, prop := range props {
+		switch prop.Name {
+		case "Rule_Description":
+			info.Description = prop.Value
+		case "Check_Id":
```
marcusburghardt: I think we could remove the "Check" information since it duplicates the "Rule" information. WDYT?

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,619 @@
...
+
+// renderKeyValuePair renders a single key-value pair using predefined lipgloss styles.
+func renderKeyValuePair(key, value string) string {
+	return keyStyle.Render(key) + ": " + valueStyle.Render(value)
+}
+
+// getControlRulesColumnsAndRows prepares columns and rows for a specific control's rules table.
+func getControlRulesColumnsAndRows(control Control) ([]table.Column, []table.Row) {
+	// Sort rules by ID for logical ordering in the table
+	sort.Slice(control.Rules, func(i, j int) bool {
```
marcusburghardt, line 373: A sort logic is used three times. Maybe we could create a generic function to centralize the sort approach?

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,619 @@
...
+			control.ImplemenationStatus,
+			strings.Join(removeDuplicates(plugins), ", "),
+		}
+		rows = append(rows, row)
+	}
+
+	columns := []table.Column{
+		{Title: "Control ID", Width: colWidthControlID},
+		{Title: "Control Title", Width: colWidthControlTitle},
+		{Title: "Implementation Status", Width: colWidthImplStatus},
```
marcusburghardt: Maybe we could use simply `Status` instead of `Implementation Status` to keep it shorter?

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,619 @@
...
+}
+
+// newControlInfoModel creates a Tea model for displaying specific control details.
+func newControlInfoModel(control Control, rowLimit int) terminal.Model {
+	// Prepare the header message with control details
+	wrappedDescription := terminal.WrapText(control.Description, 60)
+	headerFields := strings.Join([]string{
+		renderKeyValuePair("Control ID", control.ID),
+		renderKeyValuePair("Title", control.Title),
+		renderKeyValuePair("Implementation Status", control.ImplemenationStatus),
```
marcusburghardt: A shorter text could be better. Maybe only "Status"?

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,619 @@
...
+}
+
+// newRuleInfoModel creates a Bubble Tea model for displaying specific rule details.
+func newRuleInfoModel(ruleDetails RuleInfo, setParameters map[string][]string, rowLimit int) terminal.Model {
+
+	headerFields := strings.Join([]string{
+		renderKeyValuePair("Rule ID", ruleDetails.ID),
+		renderKeyValuePair("Rule Description", ruleDetails.Description),
+		renderKeyValuePair("Check ID", ruleDetails.CheckID),
+		renderKeyValuePair("Check Description", ruleDetails.CheckDescription),
```
marcusburghardt: Currently, these two lines duplicate the information from the previous lines. We can remove them. If necessary in the future, we can implement. WDYT?

marcusburghardt (commented):
I know it is in draft and changes are still expected, but I shared some comments from an early review. There are also some warnings apparently related to content being showed before the tables, but this I am also assuming is expected as it is in Draft. Besides that, the functionality was working fine with my tests using complytime-demos. I am looking forward to see the how it will evolve.

cmd/complytime/cli/info.go
gvauter: The challenge here would be passing the desired sort key as it is not consistent between the different implementation.  E.g. some sort by the `.ID` attribute, another is simply sorting strings.  The centralized function might end up more complicated than just calling sort.Slice?  WDYT?

cmd/complytime/cli/info.go
gvauter: @marcusburghardt - yes this makes sense, thank you.  I've added new logic to show all rows unless the user passes a `--limit` flag.

cmd/complytime/cli/info.go
gvauter: @marcusburghardt - yes good call, there is duplication going on between the two structs.  I will fix, thank you.

cmd/complytime/cli/info.go
marcusburghardt: I agree with your argument George. If the effort would be bigger than the final value, all fine. I also though more and don't expect changes that could demand these sort lines to be updated.

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,644 @@
...
+	)
+
+	tbl.SetStyles(table.Styles{
+		Header: tableHeaderStyle,
+		Cell:   tableCellStyle,
+	})
+
+	return terminal.Model{
+		Table:   tbl,
+		HelpMsg: fmt.Sprintf("Showing %d of %d available contorls . Use --limit to limit table rows.", tableHeight-1, len(rows)),
```
marcusburghardt:
```suggestion
		HelpMsg: fmt.Sprintf("Showing %d of %d available controls. Use --limit to limit table rows.", tableHeight-1, len(rows)),
```

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,644 @@
...
+		PreRun: func(cmd *cobra.Command, args []string) {
+			if len(args) == 1 {
+				infoOpts.complyTimeOpts.FrameworkID = filepath.Clean(args[0])
+			}
+		},
+		RunE: func(_ *cobra.Command, _ []string) error { return runInfo(infoOpts) },
+	}
+	cmd.Flags().StringVarP(&infoOpts.controlID, "control", "c", "", "show info for a specific control ID")
+	cmd.Flags().StringVarP(&infoOpts.ruleID, "rule", "r", "", "show info for a specific rule ID")
+	cmd.Flags().IntVarP(&infoOpts.limit, "limit", "l", 0, fmt.Sprintf("limit the number of table rows"))
```
marcusburghardt:
```suggestion
	cmd.Flags().IntVarP(&infoOpts.limit, "limit", "l", 0, "limit the number of table rows")
```
Also flagged in lint test.

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,644 @@
...
+
+	// Display info based on controlID or ruleID flag being passed at CLI
+	if opts.controlID != "" {
+		return displayControlInfo(opts, controlMap)
+	} else if opts.ruleID != "" {
+		return displayRuleInfo(opts, opts.ruleID, ruleRemarksMap, remarksPropsMap, setParameters)
+	} else {
+		return displayAllControls(opts, controlMap)
+	}
+
```
marcusburghardt, line 167:
```suggestion
```
Super nit. : )

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,644 @@
...
+	Parameters  []string
+}
+
+// Control repsents details about a control across
+// component sources.
+type Control struct {
+	ID                  string
+	Title               string
+	Description         string
+	ImplemenationStatus string
```
marcusburghardt:
```suggestion
	ImplementationStatus string
```
Unfortunately the typo was propagated. Could you update it, please?

marcusburghardt (changes requested):
Thanks @gvauter , the functionality is good. I only have some minor comments.

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,643 @@
...
+}
+
+func infoCmd(common *option.Common) *cobra.Command {
+	infoOpts := &infoOptions{
+		Common:         common,
+		complyTimeOpts: &option.ComplyTime{},
+	}
+	cmd := &cobra.Command{
+		Use:     "info <framework-id> [flags]",
+		Short:   "Import OSCAL data from API",
```
jpower432: Is this description right? Seems like it might be describing another command.

marcusburghardt (dismissed):
LGTM

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,643 @@
...
+	}
+	return validationComponents
+}
+
+// extractRuleDetails parses properties to fill a RuleInfo struct.
+func extractRuleDetails(props []oscalTypes.Property) Rule {
+	info := Rule{}
+	for _, prop := range props {
+		switch prop.Name {
+		case "Rule_Description":
```
jpower432, line 333: Nit: There are constants defined in `oscal-sdk-go` with these values

jpower432 (changes requested):

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,643 @@
...
+					return control.Title, nil
+				}
+			}
+		}
+	}
+	return "", fmt.Errorf("title for control '%s' not found in catalog", controlID)
+}
+
+// filterFrameworkComponents filters component definitions by framework ID.
+func filterFrameworkComponents(componentDefinitions []oscalTypes.ComponentDefinition, frameworkID string) []oscalTypes.DefinedComponent {
```
jpower432: I think the discovery of frameworks here is done differently than the list command leading to unexpected results during testing. In upstream representations of th validation component, it does not have a control implementation - https://github.com/oscal-compass/e2e-demo-cd/blob/develop/component-definitions/oscap/component-definition.json.

jpower432 (changes requested):

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,673 @@
...
+	infoContainerStyle = lipgloss.NewStyle().
+				Border(lipgloss.NormalBorder()).
+				BorderForeground(lipgloss.Color("240")). // Light grey border for containers
+				Padding(0, 1).
+				Width(90) // Fixed width for consistent formatting
+)
+
+// Rule represents details about a rule for easy
+// mapping of rules to plugins.
+type Rule struct {
```
jpower432: Nit: Curious why these are exported types. Do they need to be?

jpower432 (dismissed):
LGTM. I left one nit, but not blocking.

cmd/complytime/cli/info.go
gvauter: @jpower432 no they do not, so I will push an update.  I originally had them in `internal` but then decided to keep inside the command.  This was an oversight of the copy/paste on my part.  Thanks.

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,673 @@
...
+// component sources.
+type control struct {
+	ID                   string
+	Title                string
+	Description          string
+	ImplementationStatus string
+	Rules                []rule
+}
+
+// RulePluginMap maps a Rule ID to the of plugin that implement it.
```
marcusburghardt:
```suggestion
// RulePluginMap maps a Rule ID to the plugin that implement it.
```
Nit

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,673 @@
...
+	// Container style for key-value information blocks
+	infoContainerStyle = lipgloss.NewStyle().
+				Border(lipgloss.NormalBorder()).
+				BorderForeground(lipgloss.Color("240")). // Light grey border for containers
+				Padding(0, 1).
+				Width(90) // Fixed width for consistent formatting
+)
+
+// rule represents details about a rule for easy
+// mapping of rules to plugins.
```
marcusburghardt:
```suggestion
// rule represents details about a rule for easy mapping of rules to plugins.
```
Nit: The line is not so long. Most of the comments are using one line.

cmd/complytime/cli/info.go
```diff
@@ -0,0 +1,673 @@
...
+type rule struct {
+	ID          string
+	Plugin      string
+	Description string
+	Parameters  []string
+}
+
+// control repsents details about a control across
+// component sources.
```
marcusburghardt:
```suggestion
// control repsents details about a control across component sources.
```
Nit: The line is not so long. Most of the comments are using one line.

marcusburghardt (changes requested):
I have only aesthetics comments. Also, could you review the titles in the last two commits, please? The last is too short, and the one before is too long. : ) Maybe part of it could be moved to the commit description?

cmd/complytime/cli/info.go
gvauter: @marcusburghardt - I have fixed these issues, thank you for the feedback.

jpower432 (approved):
LGTM

marcusburghardt (approved):
Thanks @gvauter

PR #134, opened 2025-05-22
feat: add assessment filter logic CPLYTM-237 CPLYTM-760 CPLYTM-758 CPLYTM-757

jpower432:
## Summary

This PR encompasses all commits relating to the workspace Assessment Plan customization through control filtering.

## Related Issues

Closes to CPLYTM-237
Relates #127 
Relates to #132 

## Review Hints

- Review each commit individually
- To test, see the `planExample`

```bash
complytime plan myframework
complytime plan myframework --dry-run > config.yml
complytime plan myframework --scope-config config.yml
```


cmd/complytime/cli/plan.go
```diff
@@ -94,3 +153,23 @@ func loadPlan(opts *option.ComplyTime, validator validation.Validator) (*oscalTy
...
 }
+
+// planDryRun leverages the AssessmentScope structure to populate tailoring config.
+// The config is written to stdout.
+func planDryRun(frameworkId string, cds []oscalTypes.ComponentDefinition, output string) error {
+	scope, err := plan.NewAssessmentScopeFromCDs(frameworkId, cds...)
+	if err != nil {
+		return fmt.Errorf("error creating assessment scope for %s", frameworkId)
+	}
+	data, err := yaml.Marshal(&scope)
```
marcusburghardt, line 164: Since we already use json in many packages, I am curious if the idea to use yaml here was to make it simpler for the users or is there any other reason? I ask because we could have one dependency less. Just curious on the reason.

Makefile
```diff
@@ -62,3 +70,15 @@ format:
...
 	go vet ./...
 .PHONY: vet
+
+##@ Help
+
+GREEN := \033[0;32m
+TEAL := \033[0;36m
+CLEAR := \033[0m
+
+help: ## Show this help.
```
marcusburghardt, line 82: Cool!

marcusburghardt (commented):
I saw the controls reference are removed from the assessment-plan, but the rules linked to them are kept, resulting in no changes in the policy when executing the `generate` and `scan` commands. I imagine this will be implemented later, when extending the feature for rules, correct?

Also, among these 27 commits there are some, specially the first 10, where the commit message does not reflect the code changes. Maybe we could squash some commits to make the history easier to be checked/reviewed.

cmd/complytime/cli/plan.go
jpower432: Good question @marcusburghardt. Since this configuration is expected to be manually edited, YAML was chosen for better readabilty. I think this will be helpful especially as the configuration grow in complexity.

hbraswelrh (approved):
LGTM.

cmd/complytime/cli/plan.go
marcusburghardt: Thanks for the context @jpower432 .

marcusburghardt (approved):
This PR evolved in a feature branch with multiple parts developed separately and then merged with some refactoring. This explains the 27 commits where some of them are not providing much context. But at the end, the feature is working as expected. We only probably need to squash when merging to keep the main branch clean.

jpower432:
Thanks @marcusburghardt. I will complete a rebase and then squash merge this.

PR #139, opened 2025-05-29
CPLYTM-237 fix: applies fixes to broken build and UX

jpower432:
## Summary

Updates assessment-plan logic

- Fix broken build by removing redefined `d` shorthand
- Introduces `out` flag to the user can control where the config is written to.

### Rationale on providing `out` instead of writing to the workspace

For clear differentiation for the user on what should and should not be edited. Files written to the workspace and managed by `complytime` as configuration and to store state in between command runs should not be edited. Allowing the user to provide a location for the configuation gives a clear indication that the configuration is a different use case and the intent is to modify the contents.

## Related Issues
CPLYTM-237

## Review Hints

Review each commit individually

Run below to test:
```
complytime list
complytime plan <framework-id> --dry-run
complytime plan <framework-id> --dry-run --out config.yml
complytime plan <framework-id> --with-config config.yml
```

### DIFF
--- a/cmd/complytime/cli/plan.go
+++ b/cmd/complytime/cli/plan.go
@@ -21,6 +21,9 @@ import (
 
 const assessmentPlanLocation = "assessment-plan.json"
 
+// TODO: `assessmentPlanFilterLocation` is generic, but can be input based on the user preference
+const assessmentPlanFilterLocation = "config.yml"
+
 // PlanOptions defines options for the "plan" subcommand
 type planOptions struct {
 	*option.Common
--- a/cmd/complytime/cli/plan.go
+++ b/cmd/complytime/cli/plan.go
@@ -21,9 +21,6 @@ import (
 
 const assessmentPlanLocation = "assessment-plan.json"
 
-// TODO: `assessmentPlanFilterLocation` is generic, but can be input based on the user preference
-const assessmentPlanFilterLocation = "config.yml"
-
 // PlanOptions defines options for the "plan" subcommand
 type planOptions struct {
 	*option.Common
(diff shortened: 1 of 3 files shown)


hbraswelrh (approved):
LGTM. Tested locally and this works as expected :+1: . I agree with your rationale for introducing the `--out` flag. The separation makes it clear to the user that they should only edit the configuration file indicated by passing the `--out` flag.

hbraswelrh (approved):
LGTM

cmd/complytime/cli/plan.go
```diff
@@ -34,15 +31,20 @@ type planOptions struct {
...
 
 var planExample = `
-	# The default behavior is to prepare a default assessment plan with all defined
-    # controls within the framework in scope
+	# The default behavior is to prepare a default assessment plan with all defined controls within the framework in scope
 	complytime plan myframework
 
 	# To customize the assessment plan, run in dry-run mode
-	complytime plan myframework --dry-run > config.yml
+	complytime plan myframework --dry-run
```
marcusburghardt: This command is not actually customizing the assessment plan, but showing the included controls. Maybe the comment is outdated.

cmd/complytime/cli/plan.go
```diff
@@ -31,15 +31,20 @@ type planOptions struct {
...
 	withConfig string
+
+	// Out
+	Output string
 }
 
 var planExample = `
-	# The default behavior is to prepare a default assessment plan with all defined
-    # controls within the framework in scope
+	# The default behavior is to prepare a default assessment plan with all defined controls within the framework in scope
```
marcusburghardt: I know it was already here before this PR, but since the example in cobra.Command definition is calling this variable, the indentation showed with `--help` is not aligned with other commands and there is an extra blank line. I would not block this PR because of this, but it would be nice to have the examples consistent among commands.

marcusburghardt (changes requested):
It is working as expected. There are only two comments regarding the example. One is about consistency and not blocking this PR, the other is likely an outdated comment that we should update.

marcusburghardt (approved):
Thanks for the updates @jpower432 .

PR #141, opened 2025-05-30
CPLYTM-831: remove pre-release workflow

marcusburghardt:
## Summary

This workflow is creating several pre-releases are were considered not bringing value at this moment so removing this workflow will not only make the releases cleaner but also save CI resources.

## Related Issues

Cleaner releases: https://github.com/complytime/complytime/releases 

## Review Hints

The history of these jobs can be checked here:

- https://github.com/complytime/complytime/actions/workflows/pre-release.yml


jpower432 (approved):
LGTM

hbraswelrh (approved):
LGTM.

PR #147, opened 2025-06-03
chore: update openscap observation title

qduanmu:
## Summary
The check id is added as an oscal observation property, so revert the change on observation title in openscap pvp result.


marcusburghardt:
@qduanmu , I guess we had similar issues with pre-commit. :) For information, I also updated the README.md in https://github.com/complytime/complytime/pull/148 .

qduanmu:
> @qduanmu , I guess we had similar issues with pre-commit. :) For information, I also updated the README.md in #148 .

Aha, I can rebase after that merged.

marcusburghardt (dismissed):
LGTM. Only need to rebase.

jpower432 (changes requested):
@qduanmu The version of C2P needs to be bumped to the commit with your changes to ensure the upstream fixes are incorporated.

jpower432 (approved):
LGTM. Thanks @qduanmu!

marcusburghardt (commented):
Before the PR the titles were shown like this:
`"title": "package_dnf-automatic_installed",`

Now they are like this:
`"title": "xccdf_org.ssgproject.content_rule_package_dnf-automatic_installed",`

I have the impression it was not changed before bumping the versions. The longer form is how the rule is referenced in OpenSCAP Datastream (technology specific), so I think keeping it without the prefix `xccdf_org.ssgproject.content_rule_` was intentional before. WDYT @jpower432 @gvauter ?

jpower432:
@marcusburghardt For context on the reasoning behind original change - 
The title was intentionally changed through this [commit](https://github.com/complytime/complytime/commit/2937bb685ffe21d37abae98e1f23de2aabcb2ed8), but for continued functionality due to an upstream issue. Now that it is fixed, this PR just changes it back.

However, I think you bring up an excellent point! We are now relying on a property instead of the `title` during the reporting step so there are no constraints around what the `title` can be. This could be an opportunity to agree on what makes the most sense in the field.

marcusburghardt:
> @marcusburghardt For context on the reasoning behind original change - The title was intentionally changed through this [commit](https://github.com/complytime/complytime/commit/2937bb685ffe21d37abae98e1f23de2aabcb2ed8), but for continued functionality due to an upstream issue. Now that it is fixed, this PR just changes it back.
> 
> However, I think you bring up an excellent point! We are now relying on a property instead of the `title` during the reporting step so there are no constraints around what the `title` can be. This could be an opportunity to agree on what makes the most sense in the field.

Thanks for the context @jpower432 . Understood. I am also curious on more inputs on the better experience in the field.

marcusburghardt (approved):
LGTM. I have a question in https://github.com/complytime/complytime/pull/147#issuecomment-2942977265 but understand it is not in the scope of this PR. It is something we can discuss and change later, if necessary. Thanks @qduanmu

PR #159, opened 2025-06-04
ci: include guiding comment in dependabot PRs

marcusburghardt:
## Summary

This PR includes two commits:
- Fix the checkout reference so the versions can be extracted from the commit title
- Include a comment in PRs with guidance to move forward.

## Related Issues

Better experience reviewing dependabot PRs

## Review Hints

It is better to review the commits chronologically.
This should be the final changes in this workflow. It is minimalist, but can easily be extended on demand.

### DIFF
--- a/.github/workflows/dependabot-auto-review.yml
+++ b/.github/workflows/dependabot-auto-review.yml
@@ -26,19 +26,19 @@ jobs:
         run: |
           COMMIT_SUBJECT=$(git log -1 --format=%s)
           versions=($(grep -Eo '[0-9]+\.[0-9]+\.[0-9]+' <<< "$COMMIT_SUBJECT"))
-          echo "CURRENT_VERSION=${versions[0]}" >> $GITHUB_ENV
-          echo "NEW_VERSION=${versions[1]}" >> $GITHUB_ENV
+          echo "FROM_VERSION=${versions[0]}" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot-auto-review.yml
```diff
@@ -71,3 +71,16 @@ jobs:
...
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          body: |
+            Version bump from ${{ env.FROM_VERSION }} to ${{ env.TO_VERSION }}
+            Breaking risk based on semantic versioning: ${{ env.RISK }}
+            Before merging, please ensure:
+            - The update does not introduce breaking changes.
+            - The update has passed in CI tests.
+            - If the risk is high, at least two maintainers approve.
+            - If the risk is medium or low, at least one maintainer approve.
```
huiwangredhat: One approval conflicts with the PR merge rule, which requires two approvals.

marcusburghardt: The idea is that this Workflow gives one approval for PRs with low and medium risk. But since this would still demand changes in CODEOWNERS, I think you brought a good point. I will quickly update it to leave this message more generic for now.

marcusburghardt: Since the PR is small, I amended the last commit.

huiwangredhat (approved):
Thanks for your update. LGTM.

hbraswelrh (approved):
LGTM.

PR #155, opened 2025-06-04
feat: CPLYTM-759 updating scope config command flag

hbraswelrh:
## Summary
This PR changes the `--with-config` flag to `--scope-config` to reduce confusion and overlap of configuration flags. The update emphasizes the scope that is altered based on `FrameworkId` and the excluded controls of the `config.yml`.

## Related Issues
- CPLYTM-759
- Depends on PR #134 
- Closed by #156 
## Review Hints

**Useful commands:**

```bash
./bin/complytime list
./bin/complytime plan <framework-id> --dry-run 
./bin/complytime plan <framework-id> --dry-run --out config.yml
./bin/complytime plan <framework-id> --scope-config config.yml
```

PR #140, opened 2025-05-30
feat: CPLYTM-830 filtering added controls by frameworkID

hbraswelrh:
## Summary
This PR updates the check for applicable controls based on the `frameworkID`. The `frameworkID` Property on the `ControlImplementationSet` is compared with the scoped `frameworkID` . Based on the check, the controls are included if applicable. 

## Related Issues

- CPLYTM-237
- Closed by #171 

## Review Hints

**Testing with the available frameworkID in the user workspace**
- `complytime plan <frameworkid> --dry-run --out config.yml`
-  `complytime plan <frameworkid> --scope-config config.yml`
- `cat config.yml`

PR #156, opened 2025-06-04
docs: CPLYTM-759 updating for scope-config refactor

hbraswelrh:
## Summary
This PR updates the documentation for the `complytime plan` command. The `--scope-config` flag loads the `config.yml` to configure the `assessment-plan.json`.  

## Related Issues

- Depends on PR #155 and PR #134 

## Review Hints
**Related commands:**

```bash
./bin/complytime list
./bin/complytime plan <framework-id> --dry-run 
./bin/complytime plan <framework-id> --dry-run --out config.yml
./bin/complytime plan <framework-id> --scope-config config.yml
```

PR #171, opened 2025-06-10
feat: CPLYTM-830 skipping activities with out-of-scope controls

hbraswelrh:
## Summary
This PR updates the OSCAL Assessment Plan `activities` field when filtering controls based on the assessment-plan. If the controls are excluded from the `assessment-plan.json` or deleted from the `config.yml`, the associated activities will be marked as "skipped" in the `assessment-plan.json`.

The `scope_test.go` updates were made to include a test for "All Controls Out-of-Scope" and "Some Controls Out-of-Scope." The tests will trigger the out-of-scope activities with `skippedProperty` and `skippedActivity` properties when expected `control-ids` are not provided. 

## Related Issues

- Depends on #134 
- Closes #140 
- Related to #156 
- Closes #155 
- CPLYTM-237[CPLYTM-830]

## Review Hints

**Review commit by commit** 
-   [Expanding unit tests](https://github.com/complytime/complytime/pull/171/commits/28e76490cb0033d3085562776cf0aa57e39d9e1c) 
-   [Introducing skipped property](https://github.com/complytime/complytime/pull/171/commits/8d8c72302a641bc223d0b9b13bc87b258c743863)
- Note: tests are passing in CI

#### Writing the config to the complytime workspace 
- `./bin/complytime plan <frameworkid> --dry-run`
- `./bin/complytime plan <frameworkid> --dry-run --out config.yml`
#### Update the `config.yml` to exclude r1 from the assessment-plan.json
- `./bin/complytime plan <frameworkid> --scope-config config.yml`
**Result:**
image

### DIFF
--- a/internal/complytime/plan/scope.go
+++ b/internal/complytime/plan/scope.go
@@ -3,12 +3,8 @@
 package plan
 
 import (
-	"os"
-
 	oscalTypes "github.com/defenseunicorns/go-oscal/src/types/oscal-1-1-3"
 	"github.com/hashicorp/go-hclog"
-
-	"github.com/complytime/complytime/pkg/log"
 )
 
 // AssessmentScope sets up the yaml mapping type for writing to config file.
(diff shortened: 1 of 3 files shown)


internal/complytime/plan/scope.go
```diff
@@ -54,6 +54,18 @@ func (a AssessmentScope) applyControlScope(assessmentPlan *oscalTypes.Assessment
 					for controlSelectionI := range controlSelections {
 						controlSelection := &controlSelections[controlSelectionI]
 						filterControlSelection(controlSelection, includedControls)
+						if controlSelection.IncludeControls == nil {
```
jpower432, line 57: Nit: Not really a consideration right now but this marks the `RelatedControls` as nil if one of the `ControlSelections` is nil and there could be others that are populate. A note for the future because in our use case we only ever have one `ControlSelection`.

jpower432 (approved):
LGTM. One nit (no blocking)

internal/complytime/plan/scope.go
hbraswelrh: In that case, the future option would be to iterate through `ControlSelections` and for each `ControlSelection` == nil, a slice could keep track for a nil check at the end. Then, only if `ControlSelections` is empty would the `RelatedControls` would be set to nil.

PR #172, opened 2025-06-11
chore: update spec file and Makefile

marcusburghardt:
## Summary

Align the spec file with Fedora Package Guidelines.
Also make the management of environment variables mode flexible in Makefile so we can also build outside a git environment.

## Related Issues

- Keep the spec ready to be proposed in Fedora

## Review Hints

This PR also aligns the SPEC file with Fedora Package Guidelines:
- https://docs.fedoraproject.org/en-US/packaging-guidelines/
- https://docs.fedoraproject.org/en-US/packaging-guidelines/Golang/


Makefile
```diff
@@ -1,12 +1,14 @@
 GO_BUILD_PACKAGES := ./cmd/...
 GO_BUILD_BINDIR :=./bin
-GIT_COMMIT := $(or $(SOURCE_GIT_COMMIT),$(shell git rev-parse --short HEAD))
-GIT_TAG :="$(shell git tag | sort -V | tail -1)"
+GIT_COMMIT ?= $(shell git rev-parse --short HEAD 2>/dev/null || echo "unknown")
```
jpower432, line 3: No issues with the Makefile changes. I think there is still some discussion around the spec file approach, right?

marcusburghardt: Right @jpower432 . Although this spec file works in alignment to Fedora Package Guidelines and can be built using packit locally, I expect some updates regarding the dependencies and possibly some flags. But they will be more related to how we build complytime instead of spec style.

jpower432 (approved):
LGTM

AlexXuan233 (approved):
LGTM

PR #173, opened 2025-06-11
CPLYTM-843 fix: removes control duplicates from config

jpower432:
## Summary

 Removes control duplicates from config

## Related Issues
- Closes CPLYTM-843

## Review Hints

- Review each commit individually
- To test, see the `planExample`

### Testing Steps

```bash
complytime plan myframework
complytime plan myframework --dry-run > config.yml
complytime plan myframework --scope-config config.yml
```

Ensure the `config.yml` does not have the same control twice.


gvauter (approved):
Tested locally with complytime-demo, LGTM.  Just one thought on the UX of `--out`, since `--dry-run` without `--out` is already printing to stdout, should perhaps the default of `--out` be a flat file (eg: config.yml) instead of `-`?

hbraswelrh (approved):
LGTM.

jpower432:
Thanks @gvauter - `dry-run` is just a boolean value that controls whether the assessment-plan is written out or the config is written out. When `dry-run` is set, `out` controls where the config is written out. `out` is set to `-` indicating the default location is stdout.

PR #176, opened 2025-06-13
feat: CPLYTM-817 CPLYTM-818 apply customized plugin configuration

qduanmu:
## Summary
_With this change, user could apply customized plugin options via drop-in configuration file._

## Related Issues

- _Closes CPLYTM-817_

## Review Hints

- _Create a plugin configuration json file with options specified in plugin manifest, then add its directory as the value of withPluginConfig of commands._
```
$ complytime generate -c /etc/complytime/config.d --debug
$ complytime scan -c /etc/complytime/config.d
```

### DIFF
--- a/internal/complytime/plugins.go
+++ b/internal/complytime/plugins.go
@@ -8,8 +8,8 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"reflect"
 
+	"github.com/hashicorp/go-hclog"
 	"github.com/oscal-compass/compliance-to-policy-go/v2/framework"
 	"github.com/oscal-compass/compliance-to-policy-go/v2/framework/actions"
 	"github.com/oscal-compass/compliance-to-policy-go/v2/plugin"
@@ -56,22 +56,17 @@ func (p PluginOptions) Validate() error {
 
 // ToMap transforms the PluginOption struct into a map that can be consumed
 // by the C2P Plugin Manager.
-func (p PluginOptions) ToMap(pluginId string) (map[string]string, error) {
+func (p PluginOptions) ToMap(pluginId string, logger hclog.Logger) (map[string]string, error) {
 	selections := make(map[string]string)
-	val := reflect.ValueOf(p)
-	t := val.Type()
-	for i := 0; i < val.NumField(); i++ {
-		field := val.Field(i)
-		fieldType := t.Field(i)
-		key := fieldType.Tag.Get("config")
-		selections[key] = field.String()
-	}
-	if selections["userconfigroot"] != "" {
-		configPath := filepath.Join(selections["userconfigroot"], "c2p-"+pluginId+"-manifest.json")
-		delete(selections, "userconfigroot")
+	selections["workspace"] = p.Workspace
+	selections["profile"] = p.Profile
+
+	if p.UserConfigRoot != "" {
+		configPath := filepath.Join(p.UserConfigRoot, "c2p-"+pluginId+"-manifest.json")
 		configFile, err := os.Open(configPath)
 		if err != nil {
 			if os.IsNotExist(err) {
+				logger.Debug(fmt.Sprintf("Plugin manifest file does not exist: %s", configPath))
 				return selections, nil
 			}
 			return selections, fmt.Errorf("failed to open plugin config file: %w", err)
@@ -88,17 +83,20 @@ func (p PluginOptions) ToMap(pluginId string) (map[string]string, error) {
 			if configOption.Name == "workspace" || configOption.Name == "profile" {
 				continue
 			} else {
+				if configOption.Default == nil {
+					return selections, fmt.Errorf("missing %s default in %s", configOption.Name, configPath)
+				}
 				selections[configOption.Name] = *configOption.Default
 			}
+
 		}
 	}
-	delete(selections, "userconfigroot")
 	return selections, nil
 }
 
 // Plugins launches and configures plugins with the given complytime global options. This function returns the plugin map with the
 // launched plugins, a plugin cleanup function, and an error. The cleanup function should be used if it is not nil.
-func Plugins(manager *framework.PluginManager, inputs *actions.InputContext, selections PluginOptions) (map[plugin.ID]policy.Provider, func(), error) {
+func Plugins(manager *framework.PluginManager, inputs *actions.InputContext, selections PluginOptions, logger hclog.Logger) (map[plugin.ID]policy.Provider, func(), error) {
 	manifests, err := manager.FindRequestedPlugins(inputs.RequestedProviders())
 	if err != nil {
 		return nil, nil, err
@@ -115,7 +113,7 @@ func Plugins(manager *framework.PluginManager, inputs *actions.InputContext, sel
 
 	pluginSelectionsMap := make(map[plugin.ID]map[string]string)
 	for pluginId := range manifests {
-		selectionsMap, err := selections.ToMap(pluginId.String())
+		selectionsMap, err := selections.ToMap(pluginId.String(), logger)
 		if err != nil {
 			return nil, nil, err
 		}
--- a/internal/complytime/plugins.go
+++ b/internal/complytime/plugins.go
@@ -84,6 +84,10 @@ func (p PluginOptions) ToMap(pluginId string, logger hclog.Logger) (map[string]s
 				continue
 			} else {
 				if configOption.Default == nil {
+					if pluginId == "openscap" && configOption.Name == "datastream" {
+						selections[configOption.Name] = ""
+						continue
+					}
 					return selections, fmt.Errorf("missing %s default in %s", configOption.Name, configPath)
 				}
 				selections[configOption.Name] = *configOption.Default
--- a/internal/complytime/plugins.go
+++ b/internal/complytime/plugins.go
@@ -3,8 +3,11 @@
 package complytime
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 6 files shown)


cmd/complytime/option/common.go
```diff
@@ -39,18 +40,21 @@ type ComplyTime struct {
...
 	// It is set by workspace state or command positional arguments.
-	FrameworkID string
+	FrameworkID      string
+	UserPluginConfig string
 }
 
 // BindFlags populate ComplyTime options from user-specified flags.
 func (o *ComplyTime) BindFlags(fs *pflag.FlagSet) {
 	fs.StringVarP(&o.UserWorkspace, "workspace", "w", "./complytime", "workspace to use for artifact generation")
+	fs.StringVarP(&o.UserPluginConfig, "plugin-config", "u", "", "Directory where user customized plugin manifests located.")
```
marcusburghardt: The `shorthand` doesn't seem much intuitive. Maybe `c`?

cmd/complytime/option/common.go
```diff
@@ -39,18 +40,21 @@ type ComplyTime struct {
...
 }
 
 // ToPluginOptions returns global PluginOptions based on complytime Options.
 func (o *ComplyTime) ToPluginOptions() complytime.PluginOptions {
 	pluginOptions := complytime.NewPluginOptions()
 	pluginOptions.Workspace = filepath.Clean(o.UserWorkspace)
 	pluginOptions.Profile = o.FrameworkID
+	pluginOptions.UserConfigRoot = strings.TrimSpace(o.UserPluginConfig)
```
marcusburghardt: Maybe we can use `filepath.Clean` instead since it is expected a filepath here.

internal/complytime/plugins.go
```diff
@@ -58,6 +69,34 @@ func (p PluginOptions) ToMap() map[string]string {
...
+	defer configFile.Close()
+
+	jsonParser := json.NewDecoder(configFile)
+	var configManifest plugin.Manifest
+	err = jsonParser.Decode(&configManifest)
+	if err != nil {
+		return fmt.Errorf("failed to parse plugin config file: %w", err)
+	}
+	// Override manifest options with customized configuration
+	for _, configOption := range configManifest.Configuration {
```
jpower432: I am not sure this logic is needed, C2P handles reconciling the manifest options with the [selections upstream](https://github.com/oscal-compass/compliance-to-policy-go/blob/929b7bb27d969cbf5347d6e8d61a05876709b364/plugin/manifest.go#L74). Could this be updated to populate `getSelections` function instead - https://github.com/qduanmu/complytime/blob/1a18473b7da1276b27afad60960bafdec9eea1af/internal/complytime/plugins.go#L118. Right now `getSelections` is just returning the same global options for each plugin, but it could be updated to return a plugin specific configMap.

internal/complytime/plugins.go
```diff
@@ -22,6 +25,9 @@ type PluginOptions struct {
 	// Profile is the compliance profile that the plugin should use for
 	// pre-defined policy groups.
 	Profile string `config:"profile"`
+	// UserConfigRoot is the root directory where users customize
+	// plugin configuration options
+	UserConfigRoot string `config:"userconfigroot"`
```
jpower432, line 30: Wondering about the include of this in the GlobalOptions. Do you see a use case where I plugin would need access to the information in the user config directory?

jpower432 (changes requested):
I tested this locally and confirmed everything worked well. Just a few questions about the implementation.

internal/complytime/plugins.go
marcusburghardt: The idea to make it global was to make the implementation simpler instead of repeating it in each command. I was reviewing and testing the PR now and saw that other commands are also extending the options locally. So, maybe we should follow this approach also for `generate` and `scan`, which are currently the only two commands that will interact with plugins. It was already this way when initially proposed in the draft by @qduanmu and I suggested her to change. :(

marcusburghardt (changes requested):
My comments more about details, but there is a good points from @jpower432

cmd/complytime/option/common.go
qduanmu: If  `filepath.Clean` is used, it will return the '.'  for ""(the option default value ) here.

internal/complytime/plugins.go
```diff
@@ -40,12 +46,17 @@ func (p PluginOptions) Validate() error {
...
+			return errors.New("user config root does not exist")
+		}
+	}
 	return nil
 }
 
 // ToMap transforms the PluginOption struct into a map that can be consumed
 // by the C2P Plugin Manager.
-func (p PluginOptions) ToMap() map[string]string {
+func (p PluginOptions) ToMap(pluginId string) (map[string]string, error) {
```
jpower432: The changes look good here. Thanks @qduanmu! I do think we can simplify the logic a bit more and remove the reflection. I think the `PluginOptions` struct is pretty stable and we can trade the extensibilty for clarify. Do you agree @marcusburghardt?

Something like (I tested this locally):
```
func (p PluginOptions) ToMap(pluginId string) (map[string]string, error) {
	selections := make(map[string]string)
	selections["workspace"] = p.Workspace
	selections["profile"] = p.Profile

	if p.UserConfigRoot != "" {
		configPath := filepath.Join(p.UserConfigRoot, "c2p-"+pluginId+"-manifest.json")
		configFile, err := os.Open(configPath)
		if err != nil {
			if os.IsNotExist(err) {
				// Probably good to log a debug message
				return selections, nil
			}
			return selections, fmt.Errorf("failed to open plugin config file: %w", err)
		}
		defer configFile.Close()

		jsonParser := json.NewDecoder(configFile)
		var configManifest plugin.Manifest
		err = jsonParser.Decode(&configManifest)
		if err != nil {
			return selections, fmt.Errorf("failed to parse plugin config file: %w", err)
		}
		for _, configOption := range configManifest.Configuration {
			if configOption.Name == "workspace" || configOption.Name == "profile" {
				continue
			} else {
				selections[configOption.Name] = *configOption.Default
			}
		}
	}
	return selections, nil
}
```

internal/complytime/plugins.go
```diff
@@ -55,7 +66,34 @@ func (p PluginOptions) ToMap() map[string]string {
 		key := fieldType.Tag.Get("config")
 		selections[key] = field.String()
 	}
-	return selections
+	if selections["userconfigroot"] != "" {
+		configPath := filepath.Join(selections["userconfigroot"], "c2p-"+pluginId+"-manifest.json")
+		delete(selections, "userconfigroot")
+		configFile, err := os.Open(configPath)
+		if err != nil {
+			if os.IsNotExist(err) {
```
jpower432, line 68: Nit: A debug message here might be useful

internal/complytime/plugins.go
```diff
@@ -55,7 +66,34 @@ func (p PluginOptions) ToMap() map[string]string {
...
+		var configManifest plugin.Manifest
+		err = jsonParser.Decode(&configManifest)
+		if err != nil {
+			return selections, fmt.Errorf("failed to parse plugin config file: %w", err)
+		}
+		for _, configOption := range configManifest.Configuration {
+			if configOption.Name == "workspace" || configOption.Name == "profile" {
+				continue
+			} else {
+				selections[configOption.Name] = *configOption.Default
```
jpower432, line 93: I recommend adding a nil check here and return and error to the user know they have an invalid config. If the default field is not there, this will panic.

jpower432 (changes requested):
Thanks for adding the changes @qduanmu. I tested locally and everything is still functioning correctly. Added some feedback on reducing complexity in one function with the introduction of this feature.

jpower432 (dismissed):
LGTM. Thanks @qduanmu!

internal/complytime/plugins.go
```diff
@@ -40,41 +46,82 @@ func (p PluginOptions) Validate() error {
...
+		var configManifest plugin.Manifest
+		err = jsonParser.Decode(&configManifest)
+		if err != nil {
+			return selections, fmt.Errorf("failed to parse plugin config file: %w", err)
+		}
+		for _, configOption := range configManifest.Configuration {
+			if configOption.Name == "workspace" || configOption.Name == "profile" {
+				continue
+			} else {
+				if configOption.Default == nil {
```
marcusburghardt, line 86: The `Datastream` can be nil since there is a mechanism to detect it dynamically based on system properties.
So, if we create a drop-in file without a `Default` definition for `Datastream`, this error appears:
```
ERROR error running complytime: errors launching plugins: missing datastream default in /etc/complytime/config.d/c2p-openscap-manifest.json
```

marcusburghardt (changes requested):
There is just one issue with the Default option for Datastream, which could be nil because the plugin can detect it dynamically later.

marcusburghardt (approved):
Thanks for the work and quick updates on this PR @qduanmu

internal/complytime/plugins.go
```diff
@@ -40,41 +46,86 @@ func (p PluginOptions) Validate() error {
...
+					if pluginId == "openscap" && configOption.Name == "datastream" {
```
jpower432, line 87: Non-blocking comment: I think we should probably add a task to review alternatives for this type of logic so we are not including plugin specific implementation details in complytime core. This could lead to issues down the road with decoupling the openscap plugin from complytime.

jpower432 (approved):
LGTM. Thanks @qduanmu. I added one comment for something that can be reviewed more thorougly in a follow-on PR.

PR #181, opened 2025-06-24
chore: CPLYTM-871 set nil default to empty string for optional configuration

qduanmu:
## Summary
_Set default for optional plugin configuration with nil default._

## Related Issues

- _Closes CPLYTM-871_

### DIFF
--- a/internal/complytime/plugins.go
+++ b/internal/complytime/plugins.go
@@ -84,11 +84,13 @@ func (p PluginOptions) ToMap(pluginId string, logger hclog.Logger) (map[string]s
 				continue
 			} else {
 				if configOption.Default == nil {
-					if pluginId == "openscap" && configOption.Name == "datastream" {
+					if configOption.Required {
+						return selections, fmt.Errorf("missing default value for required option %s in %s", configOption.Name, configPath)
+					} else {
+						logger.Warn(fmt.Sprintf("Missing default value for %s in %s, it will be set to an empty string", configOption.Name, configPath))
 						selections[configOption.Name] = ""
 						continue
 					}
-					return selections, fmt.Errorf("missing %s default in %s", configOption.Name, configPath)
 				}
 				selections[configOption.Name] = *configOption.Default
 			}


internal/complytime/plugins.go
```diff
@@ -84,11 +84,13 @@ func (p PluginOptions) ToMap(pluginId string, logger hclog.Logger) (map[string]s
 				continue
 			} else {
 				if configOption.Default == nil {
-					if pluginId == "openscap" && configOption.Name == "datastream" {
+					if configOption.Required {
+						return selections, fmt.Errorf("missing required %s default in %s", configOption.Name, configPath)
+					} else {
+						logger.Warn(fmt.Sprintf("Missing %s default in %s, set it to empty string", configOption.Name, configPath))
```
marcusburghardt:
```suggestion
						logger.Warn(fmt.Sprintf("Missing default value for %s in %s, it will be set to an empty string", configOption.Name, configPath))
```
nit: When reading, I was wondering if I, as a user, would be expected to set the empty string or if the software would to it to me. : )

internal/complytime/plugins.go
```diff
@@ -84,11 +84,13 @@ func (p PluginOptions) ToMap(pluginId string, logger hclog.Logger) (map[string]s
 				continue
 			} else {
 				if configOption.Default == nil {
-					if pluginId == "openscap" && configOption.Name == "datastream" {
+					if configOption.Required {
+						return selections, fmt.Errorf("missing required %s default in %s", configOption.Name, configPath)
```
marcusburghardt:
```suggestion
						return selections, fmt.Errorf("missing default value for required option %s in %s", configOption.Name, configPath)
```

marcusburghardt (changes requested):
It is a minor comment only about the log message.

jpower432 (commented):
LGTM. Holding approval for the log message fix and then I will approve.

marcusburghardt (approved):
Thanks @qduanmu

jpower432 (approved):
LGTM

PR #190, opened 2025-07-02
feat: CPLYTM-814 add support for excluding rules in a scope config

gvauter:
## Summary

This PR adds support for excluding rules within a scope config.  There are two ways to exclude a rule:
- per control, by specifying an `excludeRules` section under that control in the yaml
- for all controls, by specify `globalExcludeRules` as a root key in the yaml

The order of priority is: 
- `globalExcludeRules` take highest priority.  If a rule is listed here then it is excluded for all controls regardless if a control has it included (or if a control has `includeRules: "*"`
- `excludeRules` takes priority over `includeRules` within a control.  This allows the user to keep the default content of `includeRules: "*"` and simply list the few rules they want to exclude.

Note: there is effectively a 1-to-1 relationship between "rules" and "activities".  If a rule is used by more than one control, then those control IDs will be listed inside that activity. 

When a rule is "excluded" there are a few possibilities:
- if that rule is only used by a single control, the action for that rule will have a new `property` to mark it as "skipped".  
- if a rule is used by > 1 control, then the control for which the rule has been "excluded" will have it's control ID removed from the activity, but the activity will not be marked as skipped (since it is still used by at least one control)

## Review Hints

[This code block ](https://github.com/gvauter/complyctl/blob/10d669e1ffd347485420052d8373d93fd07bfaf3/internal/complytime/plan/scope.go#L190) contains logic to normalize the `includeRules`, `excludeRules` and `globalExcludeRules` so that a user does not need to explicitly specify every section.

[This loop](https://github.com/gvauter/complyctl/blob/10d669e1ffd347485420052d8373d93fd07bfaf3/internal/complytime/plan/scope.go#L356) handles the order of priority for excluding/including.  globalExcludeRules > excludeRules > includeRules.

[This loop](https://github.com/gvauter/complyctl/blob/10d669e1ffd347485420052d8373d93fd07bfaf3/internal/complytime/plan/scope.go#L210) removes the control from the activity or marks activity as skipped depending on how many "included control" remain for that activity.

Example `config.yml`:

```
frameworkId: anssi_bp28_minimal
includeControls:
- controlId: r31
  includeRules:
  - "enable_authselect" # only this will will be included
- controlId: r59
  includeRules:
  - "*" # all rules will be included except `ensure_redhat_gpgkey_installed` which is excluded globally

globalExcludeRules:
  - "ensure_redhat_gpgkey_installed" # will be excluded for all controls
```

### DIFF
--- a/internal/complytime/plan/scope.go
+++ b/internal/complytime/plan/scope.go
@@ -13,8 +13,9 @@ import (
 
 // ControlEntry represents a control in the assessment scope
 type ControlEntry struct {
-	ControlID string   `yaml:"controlId"`
-	Rules     []string `yaml:"includeRules"`
+	ControlID    string   `yaml:"controlId"`
+	IncludeRules []string `yaml:"includeRules"`
+	ExcludeRules []string `yaml:"excludeRules,omitempty"`
 }
 
 // AssessmentScope sets up the yaml mapping type for writing to config file.
@@ -25,7 +26,8 @@ type AssessmentScope struct {
 	FrameworkID string `yaml:"frameworkId"`
 	// IncludeControls defines controls that are in scope
 	// of an assessment.
-	IncludeControls []ControlEntry `yaml:"includeControls"`
+	IncludeControls    []ControlEntry `yaml:"includeControls"`
+	GlobalExcludeRules []string       `yaml:"globalExcludeRules,omitempty"`
 }
 
 // NewAssessmentScope creates an AssessmentScope struct for a given framework id.
@@ -74,8 +76,8 @@ func NewAssessmentScopeFromCDs(frameworkId string, cds ...oscalTypes.ComponentDe
 	scope.IncludeControls = make([]ControlEntry, len(controlIDs))
 	for i, id := range controlIDs {
 		scope.IncludeControls[i] = ControlEntry{
-			ControlID: id,
-			Rules:     []string{"*"}, // by default, include all rules
+			ControlID:    id,
+			IncludeRules: []string{"*"}, // by default, include all rules
 		}
 	}
 	sort.Slice(scope.IncludeControls, func(i, j int) bool {
@@ -91,6 +93,7 @@ func (a AssessmentScope) ApplyScope(assessmentPlan *oscalTypes.AssessmentPlan, l
 	// This is a thin wrapper right now, but the goal to expand to different areas
 	// of customization.
 	a.applyControlScope(assessmentPlan, logger)
+	a.applyRuleScope(assessmentPlan, logger)
 }
 
 // applyControlScope alters the AssessedControls of the given OSCAL Assessment Plan by the AssessmentScope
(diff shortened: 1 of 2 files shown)


hbraswelrh (dismissed):
LGTM. Reviewed and needs rebase

internal/complytime/plan/scope.go
```diff
@@ -198,6 +300,112 @@ func filterControlSelection(controlSelection *oscalTypes.AssessedControls, inclu
...
+				if _, exists := ruleSet[prop.Value]; !exists {
+					rules = append(rules, prop.Value)
+					ruleSet[prop.Value] = struct{}{}
+				}
+			}
+		}
+	}
+
+	// Check step-level properties for Rule_Id
+	if activity.Steps != nil {
```
jpower432: The [activity steps are mapped to check ids](https://github.com/oscal-compass/oscal-sdk-go/blob/69548bee35ccaa8f0a4352a7c15550c69fe0ca04/internal/plans/plan.go#L186) so I do not know that this is required.

internal/complytime/plan/scope.go
```diff
@@ -198,6 +300,112 @@ func filterControlSelection(controlSelection *oscalTypes.AssessedControls, inclu
...
+			}
+		}
+	}
+
+	return rules
+}
+
+// filterControlSelectionByRules removes controls from a selection if the activity's rules should be excluded for those controls
+func (a AssessmentScope) filterControlSelectionByRules(controlSelection *oscalTypes.AssessedControls, activityRules []string, controlRuleConfig map[string]ControlEntry, globalExcludeRules map[string]struct{}, logger hclog.Logger, activityTitle string) {
+	if controlSelection.IncludeControls == nil {
```
jpower432, line 275: Nit: I agree with the safeguard, but I'm thinking logically having no mapped controls for an input would be undesired. Maybe a debug log would be helpful for debugging data issues here.

internal/complytime/plan/scope.go
```diff
@@ -198,6 +300,112 @@ func filterControlSelection(controlSelection *oscalTypes.AssessedControls, inclu
 	}
 }
 
+// getActivityRules extracts rule IDs from an activity's properties and step properties
+func (a AssessmentScope) getActivityRules(activity *oscalTypes.Activity) []string {
+	var rules []string
+	ruleSet := make(map[string]struct{}) // For deduplication
+
+	// Check activity-level properties for Rule_Id
+	if activity.Props != nil {
```
jpower432: There may be some changes needed upstream to support this logic. If you generate an assessment plan with `complytime`, none of the activities has rule properties. The activity itself represents the rule so the rule ID is captured in the `activity.Title`. The goal is to move away from the title in favor of the `assessment-rule-id` and `assessment-check-id` custom properties, but that has not been implemented upstream yet.

jpower432 (changes requested):
Great work @gvauter. The logic makes sense, but there is an issue when using this with `complytime plan` the activities don't have rule id properties so `getActivityRules` returns no rules to filter. This should be checked against the activity title and changes proposed upstream to move to custom properties.

internal/complytime/plan/scope.go
gvauter: Thanks @jpower432 for the review.  I've pushed a new commit that removes this unnecessary processing of steps.  I've also removed the processing of activity props to get the rule ID and simply use activity.Title directly.

jpower432 (approved):
LGTM. Nice work!

hbraswelrh (approved):
LGTM.

PR #197, opened 2025-07-09
feat: adding controlTitle to configured assessment plan

hbraswelrh:
## Summary
This PR introduces the `controlTitle` field for awareness of control-id name in the `config.yml`.  The `controlTitle` in the `config.yml` will update based on the controls associated with the `frameworkID`. 

This PR refactored the `internal/complytime/plan` package originally located inside of `internal/complytime` for better portability and reduction of cyclic import issues. 

#### What was changed

- Refactored `internal/complytime` package to no longer include `internal/complytime/plan`, essentially flattening the `internal/complytime` package for ease of importing within the package. 
- The `GetControlTitle` function was added to `internal/complytime/plan.go` and is used in the `cli` package for getting the control titles for both the `complyctl info` and `complyctl plan` commands. 
- The `loadControlTitlesFromSource` grabs the control titles from the control source and stores them as a map to be used in the `GetControlTitle` function. 

The new structure with the compressed `internal/complytime`
```tree
internal
├── complytime
│   ├── catalogs.go
│   ├── catalogs_test.go
│   ├── configuration.go
│   ├── configuration_test.go
│   ├── controls.go
│   ├── controls_test.go
│   ├── plan.go
│   ├── plan_test.go
│   ├── plugins.go
│   ├── plugins_test.go
│   ├── scan.go
│   ├── scan_test.go
│   ├── scope.go
│   ├── scope_test.go
│   └── testdata
│       ├── complytime
│       │   ├── bundles
│       │   │   └── example-component-definition.json
│       │   ├── controls
│       │   │   ├── sample-catalog.json
│       │   │   └── sample-profile.json
│       │   └── plugins
│       │       └── c2p-openscap-manifest.json
│       └── openscap
│           ├── invalid.xml
│           └── ssg-rhel-ds.xml
├── terminal
│   ├── spinner.go
│   ├── spinner_test.go
│   ├── table.go
│   └── table_test.go
├── utils
└── version
    ├── version.go
    └── version_test.go
```
## Related Issues

- Related to PR #190 

#### Test by running the following commands
- `./bin/complyctl plan <frameworkId> --dry-run`
-  `./bin/complyctl plan <frameworkId> --dry-run --out config.yml`
-  `./bin/complyctl plan <frameworkId> --scope-config config.yml`

Alt text

> _This PR was developed with the help of Cursor AI IDE_

### DIFF
--- a/internal/complytime/plan.go
+++ b/internal/complytime/plan.go
@@ -74,17 +74,20 @@ func Settings(plan *oscalTypes.AssessmentPlan) (settings.Settings, error) {
 	return settings.Settings{}, ErrNoActivities
 }
 
-// GetControlTitle retrieves the title for a control from the catalog
-func GetControlTitle(controlID string, controlSource string, appDir ApplicationDirectory, validator validation.Validator) (string, error) {
+// loadControlTitlesFromSource loads all control titles from a source and returns them as a map
+func loadControlTitlesFromSource(controlSource string, appDir ApplicationDirectory, validator validation.Validator) (map[string]string, error) {
 	profile, err := LoadProfile(appDir, controlSource, validator)
 	if err != nil {
-		return "", fmt.Errorf("failed to load profile from source '%s': %w", controlSource, err)
+		return nil, fmt.Errorf("failed to load profile from source '%s': %w", controlSource, err)
 	}
 
 	if profile.Imports == nil {
-		return "", fmt.Errorf("profile '%s' has no imports", controlSource)
+		return nil, fmt.Errorf("profile '%s' has no imports", controlSource)
 	}
 
+	controlTitles := make(map[string]string)
+
+	// Load all control titles from all imported catalogs
 	for _, imp := range profile.Imports {
 		catalog, err := LoadCatalogSource(appDir, imp.Href, validator)
 		if err != nil {
@@ -98,11 +101,26 @@ func GetControlTitle(controlID string, controlSource string, appDir ApplicationD
 				continue
 			}
 			for _, control := range *group.Controls {
-				if control.ID == controlID && control.Title != "" {
-					return control.Title, nil
+				if control.ID != "" && control.Title != "" {
+					controlTitles[control.ID] = control.Title
 				}
 			}
 		}
 	}
+
+	return controlTitles, nil
+}
+
+// GetControlTitle retrieves the title for a specific control from the catalog
+func GetControlTitle(controlID string, controlSource string, appDir ApplicationDirectory, validator validation.Validator) (string, error) {
+	controlTitles, err := loadControlTitlesFromSource(controlSource, appDir, validator)
+	if err != nil {
+		return "", err
+	}
+
+	if title, found := controlTitles[controlID]; found {
+		return title, nil
+	}
+
 	return "", fmt.Errorf("title for control '%s' not found in catalog", controlID)
 }
--- a/internal/complytime/plan.go
+++ b/internal/complytime/plan.go
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
-package plan
+package complytime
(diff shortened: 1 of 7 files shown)


internal/complytime/scope.go
```diff
@@ -73,10 +93,16 @@ func NewAssessmentScopeFromCDs(frameworkId string, cds ...oscalTypes.ComponentDe
 	}
 
 	controlIDs := includeControls.All()
+	for _, id := range controlIDs {
```
gvauter: Perhaps this loop could be eliminated by moving the `GetControlTitle` into the loop at line 102?  Essentially build the list of includeControls, then get the title for each one as you add it to the scope.IncludeControls.

gvauter (changes requested):
One suggestoin on the additional loop.

internal/complytime/plan.go
```diff
@@ -73,3 +73,36 @@ func Settings(plan *oscalTypes.AssessmentPlan) (settings.Settings, error) {
 	}
 	return settings.Settings{}, ErrNoActivities
 }
+
+// GetControlTitle retrieves the title for a control from the catalog
+func GetControlTitle(controlID string, controlSource string, appDir ApplicationDirectory, validator validation.Validator) (string, error) {
```
jpower432: I think my feedback is related to @gvauter's, but recommend only loading the profile and the catalogs from disk once if it has been seen before and not on every control ID query which may belong to the same control implementation. Perhaps using maps able to store indexed control titles with the control source as the key. WDYT?

jpower432 (changes requested):
Thanks @hbraswelrh. Recommending a caching mechanism be used to avoid loading the same profile and catalog from disk redundantly.

internal/complytime/plan.go
hbraswelrh: @jpower432 Thank you for this helpful explanation. I agree that both you and @gvauter pointed out the inefficient loading of profiles and catalogs. I will work on mapping the `controltitles` to the `controlsource`. That sounds good. :+1:

internal/complytime/scope.go
```diff
@@ -106,7 +146,16 @@ func (a AssessmentScope) applyControlScope(assessmentPlan *oscalTypes.Assessment
 		includedControls.Add(entry.ControlID)
 	}
 	logger.Debug("Found included controls", "count", len(includedControls))
-
+	for _, controlT := range assessmentPlan.ReviewedControls.ControlSelections {
```
hbraswelrh, line 150: @gvauter @jpower432 If this is too similar of a variable to the unit test T type, I can change it to something else. I used `controlT` to represent the title.

gvauter (dismissed):
LGTM

internal/complytime/scope.go
```diff
@@ -65,6 +73,37 @@ func NewAssessmentScopeFromCDs(frameworkId string, cds ...oscalTypes.ComponentDe
...
 						if ir.ControlId != "" {
 							includeControls.Add(ir.ControlId)
+
+							// Getting control title using map lookup
+							if validator != nil {
+								if _, exists := controlTitles[ir.ControlId]; !exists {
+									// Check if we've already loaded this source
+									sourceTitles, sourceLoaded := controlTitlesBySource[ci.Source]
+									if !sourceLoaded {
+										// Load all titles from this source
```
jpower432: I think you could do this check earlier. This loop is ranging over a set of implemented requirements for the control implementation and source. What do you this about moving this to the outer loop for the control implementations from the components?

jpower432 (changes requested):
Thanks for the changes @hbraswelrh! Just left one suggestion about when the control source titles are loaded.

jpower432 (approved):
LGTM

gvauter (approved):
LGTM

PR #206, opened 2025-07-22
chore: update sample files based on CUSP for Fedora

marcusburghardt:
## Summary

Updated the sample files with more realistic content obtained from https://github.com/ComplianceAsCode/oscal-content.
The samples were based on CUSP content for Fedora.

Also updates the SPEC file to populate the complytime workspace with the samples so `complyctl` can be used out-of-box.

## Related Issues

- More realistic content to be used as sample.

## Review Hints

The component-definition file was manipulated to remove all repeated declaration of variables for each rule.
Instead, only the first rule is including all parameters. This reduce drastically the size of this component definition file and it was confirmed this had no impact on commands interacting with openscap-plugin.

Once there is a capability to map rules and variables in SSG, the oscal-content can be updated and consequently these samples files.

### DIFF
--- a/complyctl.spec
+++ b/complyctl.spec
@@ -67,6 +67,8 @@ install -d -m 0755 %{buildroot}%{_mandir}/{man1,man5,man7}
 
 # Copy sample data to be consumed by complyctl CLI
 cp -rp docs/samples %{buildroot}%{_datadir}/%{app_dir}
+install -p -m 0644 docs/samples/sample-{catalog,profile}.json %{buildroot}%{_datadir}/%{app_dir}/controls
+install -p -m 0644 docs/samples/sample-component-definition.json %{buildroot}%{_datadir}/%{app_dir}/bundles
 
 # Install files for complyctl CLI
 install -p -m 0755 bin/complyctl %{buildroot}%{_bindir}/complyctl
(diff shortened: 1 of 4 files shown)


complyctl.spec
```diff
@@ -106,6 +108,8 @@ go test -mod=vendor -race -v ./...
 %dir %{_sysconfdir}/%{app_dir}
 %dir %{_sysconfdir}/%{app_dir}/config.d
 %{_datadir}/%{app_dir}/samples/{sample-catalog.json,sample-component-definition.json,sample-profile.json,c2p-openscap-manifest.json}
+%{_datadir}/%{app_dir}/controls/{sample-catalog.json,sample-profile.json}
+%{_datadir}/%{app_dir}/bundles/sample-component-definition.json
```
qduanmu, line 112: Since these sample files will be installed into related directories here, how about removing the samples directory, WDYT?

marcusburghardt: Good observation. I also considered removing it before. The main reason to keep it is because the manifest file from samples is used in %post section for openscap-plugin. I am also not 100% sure now but it is possible we bring more samples in this directory later, so it was not removed for now.

qduanmu: Ok, let's keep them for now.

qduanmu (approved):
LGTM

huiwangredhat (approved):
LGTM.

PR #212, opened 2025-07-25
feat: CPLYTM-932 adding parameter flag to info command

hbraswelrh:
## Summary

This PR introduces the `--parameter` flag to print a styled table informing the user of the content below. This PR will be part of a larger effort to edit parameters of the assessment plan.

**Table 1:** 
* `Parameter ID`: name/id of parameter.
* `Description`: Elaborating on description of parameter.
* `Used By Rules`: Rules that use the parameter.

**Table 2:**

`Parameter Value(s)`: The possible values for the parameter (including current value)
`Status`: The **current** selection and the **alternative** selection(s)

The goal of the `complyctl info <framework-id> --parameter <parameter-id>` command is to see the available parameter alternatives. This will be coupled with a future PR that will populate fields for alternative selections in the `config.yml` for selecting which alternative to consider during assessment. 

## Related Issues

- CPLYTM-932

## Review Hints

image

* **Styled format:** `./bin/complyctl info <framework-id> --parameter <--parameter-id> --limit 5`
*  **Plain text:** `./bin/complyctl info <framework-id> --parameter <--parameter-id> --limit 5 --plain`
*  The `./bin/complyctl info <framework-id> --parameter <parameter-id>` command will find the available parameter selections associated with the `ParameterID`. The `--parameter` flag will also print the rules using the parameters. 

>  The contents of this PR were created as a _prototype_ and provided to Cursor IDE to expand on formatting and suggestions. **This PR was assisted by Cursor v1.3.9**

PR #229, opened 2025-08-13
feat: CPLYTM-900 CPLYTM-899 dry run logic

hbraswelrh:
## Summary
This PR updates the existing `--dry-run` logic for the `complyctl plan` command. The `selectParameters` YAML key was added to inform the user of the parameters that are available in their assessment plan when passing the `--dry-run` flag. The included controls of the assessment plan will have their own `selectParameters` YAML key that has `name` and `value` for each of the parameterIDs that are associated with the control ID. 

Several functions were added to maintain the `NewAssessmentScopeFromCDs` factory functionality. The smaller functions build around the `NewAssessmentScopeFromCDs` to help separate parsing and processing the control implementations. 

## Related Issues

- Related to #212 
- CPLYTM-899

## Review Hints

- Test the dry-run output with the command:  `./bin/complyctl plan <framework-id> --dry-run` 
image

- Test the output written to the `config.yml` using the command: `./bin/complyctl plan <framework-id> --dry-run --out config.yml`
image

**Example:**
* **Framework ID:** anssi_bp28_minimal

> This PR includes assisted code. The assisted code uses Cursor v 1.24.0

### DIFF
--- a/internal/complytime/scope.go
+++ b/internal/complytime/scope.go
@@ -47,6 +47,13 @@ func NewAssessmentScope(frameworkID string) AssessmentScope {
 	}
 }
 
+// componentDefResult holds the results from processing component definitions
+type componentDefResult struct {
+	includeControls   includeControlsSet
+	controlTitles     map[string]string
+	controlParameters map[string]map[string][]string
+}
+
 // NewAssessmentScopeFromCDs creates and populates an AssessmentScope struct for a given framework id and set of
 // OSCAL Component Definitions.
 func NewAssessmentScopeFromCDs(frameworkId string, appDir ApplicationDirectory, validator validation.Validator, cds ...oscalTypes.ComponentDefinition) (AssessmentScope, error) {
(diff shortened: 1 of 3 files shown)


internal/complytime/scope.go
```diff
@@ -107,18 +118,94 @@ func NewAssessmentScopeFromCDs(frameworkId string, appDir ApplicationDirectory,
...
+					if ci.SetParameters != nil {
+						// Create map of available set parameters
+						implementedSetParams := make(map[string][]string)
+						for _, sp := range *ci.SetParameters {
+							if sp.ParamId != "" && len(sp.Values) > 0 {
+								implementedSetParams[sp.ParamId] = sp.Values
+							}
+						}
+
+						remarksProps := extractRemarksProperties(cds)
```
jpower432: Is it possible any of the property parsing logic (especially for OSCAL Compass extenstions) could live upstream in `oscal-sdk-go`? It seems the `NewAssessmentScopeFromCDs` is performing a lot of manual Component Definition parsing that could be generally useful. Not a blocker for this PR, but I think might be worth looking at in a follow on from a technical debt perspetive.

internal/complytime/scope.go
```diff
@@ -107,18 +118,94 @@ func NewAssessmentScopeFromCDs(frameworkId string, appDir ApplicationDirectory,
 							}
 						}
 					}
+
+					// Process set parameters - match by remarks groups with control rules
```
jpower432: I would recommend breaking down `NewAssessmentScopeFromCDs` into smaller function for readability. Perhaps aleast moving the Component Definition parsing to a new function and make this factory function about creating the scope. WDYT?

jpower432 (commented):
Thanks @hbraswelrh. Added a couple initial comments.

internal/complytime/scope.go
hbraswelrh: I think this is a very good point. Updating `oscal-sdk-go` with the introduced property parsing logic would be beneficial for maintainability and decoupling. I could propose a task for evaluating `complyctl` and decoupling through an upstream PR. WDYT?

internal/complytime/scope.go
hbraswelrh: I broke the `NewAssessmentScopeFromCDs` into smaller functions for readability and future upstream updates in `oscal-sdk-go`.

jpower432 (approved):
LGTM

gvauter (approved):
LGTM

PR #228, opened 2025-08-13
docs: CPLYTM-898 updating the documentation for parameters

hbraswelrh:
## Summary

This PR updates the `help` example for the `complyctl plan` and `complyctl info` command. The updates to the `info` command introduced the `--parameter` flag which is documented in this change. The `selectParameters` YAML key is updated in the help example on the `README.md`.

## Related Issues

- CPLYTM-898
- Related to #222 
- Related to #231 

## Review Hints

- Updates made to man page will support this change.
- The `README.md` updates the flag references.

PR #231, opened 2025-08-14
feat: CPLYTM-900 CPLYTM-901 assessment plan regeneration logic

hbraswelrh:
## Summary

This PR updates the assessment plan regeneration logic for selecting parameters. The attached docs [regeneration/notes](https://docs.google.com/document/d/1jilafjuCaapzfqgLULUYCOcdF2Pn0x-o3xip-R0nP_A/edit?tab=t.bn9m1lr80jbt) & [notes for regeneration](https://docs.google.com/document/d/1_DakM48LiirXyowJsbI5ho_sawWP4nX0Lh4pLXkwtbg/edit?tab=t.ofaj7sp7nabz) outlines the expected output when configuring parameters of the assessment plan. The updates regenerate the assessment plan using the `remarks` values. The implemented requirements map the `control-id` to the rules remarks property that populates the associated `selectParameters`. Therefore, the control-ids with no `selectParameters` values will be marked "N/A." Otherwise, for all rules associated with a control-id, the parameter-ids that share the same remarks value will be present in the `selectParameters` field. When the user passes the `config.yml` back using the `--scope-config config.yml`,  the validation will check the implemented-requirements and find the parameter alternative values using the numeric suffix. 

Several functions were added to maintain the `NewAssessmentScopeFromCDs` factory functionality. The smaller functions build around the `NewAssessmentScopeFromCDs` to help separate parsing and processing the control implementations. 

## Related Issues

- Depends on #229 
- Related to #228 

## Review Hints

- Test by passing the `./bin/complyctl info <framework-id> --parameter <parameter-id>` to see the alternative selections for the particular `parameter-id`.
- Use the `./bin/complyctl plan <framework-id> --dry-run --out config.yml` to configure the `selectParameters` yaml key and second-level yaml keys for the parameter value alternatives seen in the info command. 
- Pass the updated `config.yml` to the `./bin/complyctl plan <framework-id> --scope-config config.yml` and the assessment plan will be regenerated with the new selected parameter values. 
- Test the error functionality by updating the `selectParameters:` `value` field with an incorrect alternative value. The result will throw an error and inform the user of the available alternative values.
image

> This PR was assisted by `Cursor v1.24.0`. The coding assistant was used to validate and expand on the use-case and tests with provided structure and logic.


internal/complytime/scope.go
```diff
@@ -65,76 +86,165 @@ func NewAssessmentScopeFromCDs(frameworkId string, appDir ApplicationDirectory,
...
+
+			for _, props := range remarksProps {
+				var ruleID string
+				var parametersInGroup []string
+
+				for _, prop := range props {
+					if prop.Name == extensions.RuleIdProp {
+						ruleID = prop.Value
+					}
+					if prop.Name == extensions.ParameterIdProp || strings.HasPrefix(prop.Name, extensions.ParameterIdProp+"_") {
```
gvauter: Just a nit, but this could be a single conditional that uses `strings.HasPrefix(prop.Name, extensions.ParameterIdProp)`.  It would cover both cases.

internal/complytime/scope.go
```diff
@@ -384,6 +652,111 @@ func (a AssessmentScope) isRuleInList(ruleID string, ruleList []string) bool {
...
+		cleanedAlternatives := removeDuplicates(allPossibleAlternatives)
+		return false, cleanedAlternatives
+	}
+
+	return true, nil
+}
+
+// isParameterIdProperty checks if a property matches Parameter_Id.
+func isParameterIdProperty(propertyName string) bool {
+	return propertyName == extensions.ParameterIdProp || strings.HasPrefix(propertyName, extensions.ParameterIdProp+"_")
```
gvauter: Same here, the "or" can be simplified to just checking for HasPrefix

gvauter (changes requested):
Two small nits.

PR #238, opened 2025-08-18
chore: [CPLYTM-941] add e2e pipeline

huiwangredhat:
## Summary
This PR is to add the pipeline for e2e. It includes the e2e pipeline workflow and initiates the test environment on Fedora container. The test cases will be included in another PR. 
## Related Issues
[CPLYTM-940](https://issues.redhat.com/browse/CPLYTM-940)
[CPLYTM-943](https://issues.redhat.com/browse/CPLYTM-943)

- _Closes #[CPLYTM-941](https://issues.redhat.com/browse/CPLYTM-941)

## Review Hints

- Install the dependencies on a Fedora container
- Build complyctl and download the contents from oscal-content to the related directories 
- Run the e2e tests(more cases will be included in another PR, the related issue is [CPLYTM-943](https://issues.redhat.com/browse/CPLYTM-943))

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,7 +1,9 @@
 name: e2e
 
 on:
-  # Triggers the workflow on pull request events for the main branch
+  push:
+    branches:
+      - main
   pull_request:
     types:
       - opened
@@ -10,51 +12,26 @@ on:
 permissions:
   contents: read
 
-jobs:
-  run-e2e-on-fedora-vm:
-    runs-on: ubuntu-latest
+env:
+  COMPLYTIME_DEV_MODE: 1
 
+jobs:
+  run-e2e:
+    runs-on: ubuntu-latest
+    container:
+      image: fedora:latest
     steps:
       # Checkout the code from the repository
       - name: Checkout code
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
       
-      # Set up Lima, the tool that will create our fedora VM.
-      - name: Set up Lima
-        uses: lima-vm/lima-actions/setup@03b96d61959e83b2c737e44162c3088e81de0886 # v1.0.1
-        id: lima
+      - name: Install the dependences
+        run: dnf install git wget make scap-security-guide go tree -y
 
-      # Use a cache for Lima to speed up subsequent runs.
-      # This is a recommended best practice to make your workflows faster.
-      - name: Cache Lima downloads
-        uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
-        with:
-          path: ~/.cache/lima
-          key: lima-${{ runner.os }}-${{ steps.lima.outputs.version }}
-
-      # Start the Fedora VM.
-      # 'limactl start' downloads the Fedora cloud image and boots it as a VM.
-      # The template://fedora shorthand makes it easy to specify.
-      - name: Start Fedora VM
-        run: limactl start --name=fedora template://fedora 
-
-      - name: Install the dependences on fedora VM
+      - name: Build complyctl and init complyctl
         run: |
-          limactl shell fedora bash -c '
-            sudo dnf install git wget make scap-security-guide go tree -y
-          '
-      - name: Build complyctl and Init complyctl
-        run: |
-          limactl shell fedora bash -c '
-            cp -R ../complyctl /tmp && cd /tmp/complyctl
-            sudo chown -R runner:runner /usr/local/
-            export COMPLYTIME_DEV_MODE=1
-            sh tests/build_init_env.sh
-            '
+          git config --global --add safe.directory /__w/complyctl/complyctl
+          sh tests/build_init_env.sh
+
       - name: Run e2e tests
-        run: |
-          limactl shell fedora bash -c '
-            cd /tmp/complyctl
-            export COMPLYTIME_DEV_MODE=1
-            go test tests/e2e/e2e_test.go -v
-            '
+        run: go test tests/e2e/e2e_test.go -v
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,60 @@
+name: e2e
+
+on:
+  # Triggers the workflow on pull request events for the main branch
+  pull_request:
+    types:
+      - opened
+      - synchronize 
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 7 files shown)


AlexXuan233:
Just a suggestion, if using container inside the workflow is possible, we better use container. Because container is much faster than vm and simpler. But container do have some limitation, so if container do not meet e2e requirement, we can use vm.
```yaml
    container:
      image: fedora:latest
```

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,60 @@
+name: e2e
+
+on:
```
AlexXuan233:
```suggestion
on:
  push:
    branches:
      - main
  pull_request:
```

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,60 @@
...
+  contents: read
+
+jobs:
+  run-e2e-on-fedora-vm:
+    runs-on: ubuntu-latest
+
+    steps:
+      # Checkout the code from the repository
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
AlexXuan233: Use `actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0` align with other workflow.

tests/build_init_env.sh
```diff
@@ -0,0 +1,37 @@
...
+# Download fedora cusp OSCAL content
+wget $URL/profiles/$PROFILE/profile.json -O $HOME/$WDIR/controls/profile.json
+wget $URL/catalogs/$CATALOG/catalog.json -O $HOME/$WDIR/controls/catalog.json
+wget $URL/component-definitions/$PRODUCT/$PROFILE/component-definition.json -O $HOME/$WDIR/bundles/component-definition.json
+# Update trestle path
+sed -i "s|trestle://catalogs/$CATALOG/catalog.json|trestle://controls/catalog.json|" $HOME/$WDIR/controls/profile.json
+sed -i "s|trestle://profiles/$PROFILE/profile.json|trestle://controls/profile.json|" $HOME/$WDIR/bundles/component-definition.json
+# Setup plugin
+cp -rp bin/openscap-plugin $HOME/$WDIR/plugins
+checksum=$(sha256sum $HOME/$WDIR/plugins/openscap-plugin| cut -d " " -f 1 )
```
AlexXuan233, line 42: `checksum` not be used.

huiwangredhat:
> Just a suggestion, if using container inside the workflow is possible, we better use container. Because container is much faster than vm and simpler. But container do have some limitation, so if container do not meet e2e requirement, we can use vm.
> 
> ```yaml
>     container:
>       image: fedora:latest
> ```

@AlexXuan233 thanks. Ok, I will update it to run the testing in a container for performance. I forgot the specific reason why we used the VM to do the demo at that moment.  If we meet any limitations in the future for a container, it is easy to update to the container to VM.

AlexXuan233:
> > Just a suggestion, if using container inside the workflow is possible, we better use container. Because container is much faster than vm and simpler. But container do have some limitation, so if container do not meet e2e requirement, we can use vm.
> > ```yaml
> >     container:
> >       image: fedora:latest
> > ```
> 
> @AlexXuan233 thanks. Ok, I will update it to run the testing in a container for performance. I forgot the specific reason why we used the VM to do the demo at that moment. If we meet any limitations in the future for a container, it is easy to update to the container to VM.

For container, there are some limitation like can not use `systemctl`, will impact the `scan` command, but we can choose rules can be performed on container. Like rule `accounts_umask_etc_login_defs` in control `cusp_fedora_4-1`
```yaml
frameworkId: cusp_fedora
includeControls:
- controlId: cusp_fedora_4-1
  controlTitle: Account Protection
  includeRules:
  - "accounts_umask_etc_login_defs"
```

tests/build_init_env.sh
huiwangredhat: Good catch! Updated.

huiwangredhat:
@qduanmu, would you like to review this PR when you are available? I guess @marcusburghardt has been busy recently. Thank you.

marcusburghardt (commented):
LGTM, but some changes are expected after merging https://github.com/complytime/complyctl/pull/239 and rebasing so we can unify the way we prepare the testing environment.

huiwangredhat:
@AlexXuan233 @marcusburghardt I rebased the PR 239, and did some updates. Would you like to review it again? Thank you.

.github/workflows/unit_test.yml
```diff
@@ -30,4 +30,4 @@ jobs:
 
       # Step 4: Run Go tests
       - name: Run unit tests
-        run: go test -v ./...
+        run: go test -v $(go list ./... | grep -v /tests)
```
marcusburghardt, line 33: I didn't see a difference in the output with this change. Maybe I am missing something, but is the change necessary? Or is the idea to prevent new specific tests introduced in the `/tests` to be executed automatically?

tests/build_init_env.sh
```diff
@@ -0,0 +1,45 @@
...
+    exit 1
+fi
+
+# Build the compyctl
+make build && cp ./bin/complyctl /usr/local/bin
+# Running list command that will fail due to no requirements files
+echo "Attempting to run the command complyctl list."
+set +e
+complyctl list 2>/dev/null
+echo "An error occurred, but script continues after the complyctl list."
```
marcusburghardt: Maybe we should include that the error is expected because there is no content?

marcusburghardt (dismissed):
I have some comments that don't block this PR, specially knowing that other PRs are expected to include test cases, so these comments could be addressed in future PRs if preferred.

.github/workflows/unit_test.yml
huiwangredhat: That's a good question. Right, it won't impact the e2e and integration test. It impacts the unit-test. If it won't exclude the `/tests` directory, the `unit-test` workflow will run the tests in `test/e2e/e2e_test.go` and the unit-test will failed.

tests/build_init_env.sh
huiwangredhat: Yes, that's a navigate test case. If we want to add this case to the test cases, I have no good idea how to add the setup step in the workflow :).  Later, if we have a good idea or suggestion, we could update this part. WDYT?

AlexXuan233 (approved):
LGTM

tests/build_init_env.sh
huiwangredhat: I just updated the echo message to explain the error here.

marcusburghardt (approved):
LGTM. Thanks @huiwangredhat

PR #267, opened 2025-09-10
feat: CPLYTM-983 add support for waived rules

qduanmu:
## Summary
Provide user the ability to set waived rules, so that when they failed, they could be showed as waived instead of failed in the report. 

## Related Issues

- _Closes CPLYTM-983_
- https://github.com/oscal-compass/compliance-to-policy-go/issues/230

## Review Hints
Two yaml keys could be used in the scope configuration file, 'globalWaiveRules' for global waive rules and 'waiveRules' for control specific waive rules. The priority is global > control specific, exclude > waive > include.

### DIFF
--- a/internal/complytime/scope.go
+++ b/internal/complytime/scope.go
@@ -25,6 +25,7 @@ type ControlEntry struct {
 	ControlTitle     string           `yaml:"controlTitle"`
 	IncludeRules     []string         `yaml:"includeRules"`
 	ExcludeRules     []string         `yaml:"excludeRules,omitempty"`
+	WaiveRules       []string         `yaml:"waiveRules,omitempty"`
 	SelectParameters []ParameterEntry `yaml:"selectParameters,omitempty"`
 }
 
@@ -38,6 +39,7 @@ type AssessmentScope struct {
 	// of an assessment.
 	IncludeControls    []ControlEntry `yaml:"includeControls"`
 	GlobalExcludeRules []string       `yaml:"globalExcludeRules,omitempty"`
+	GlobalWaiveRules   []string       `yaml:"globalWaiveRules,omitempty"`
 }
 
 // NewAssessmentScope creates an AssessmentScope struct for a given framework id.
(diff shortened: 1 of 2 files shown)


hbraswelrh (commented):
After adding unit tests this PR will be approved from my side. 

I tested the `waiveRules`, `globalWaiveRules`, and tested using both `excludeRules` to produce the `skipped` property at the same time of the `waived` property. All working as expected and the source code follows the structure of past implementations with `rules`. 

I would suggest mentioning Issue #[230](https://github.com/oscal-compass/compliance-to-policy-go/issues/230) in your PR body for updates impacting the upstream `compliance-to-policy-go`. 

**In the `config.yml` for r59**
```yaml
 waiveRules:
  - "ensure_gpgcheck_never_disabled"
```
**Resulting `assessment-plan.json` fields for r59**
```json
  {
       "name": "waived",
       "ns": "https://oscal-compass.github.io/compliance-trestle/schemas/oscal",
       "value": "true"
      }
```

hbraswelrh (approved):
LGTM. I tested locally and this update works as expected.

internal/complytime/scope.go
```diff
@@ -391,7 +420,21 @@ func (a AssessmentScope) applyRuleScope(assessmentPlan *oscalTypes.AssessmentPla
 									Ns:    extensions.TrestleNameSpace,
 								}
 								*step.Props = append(*step.Props, skipped)
+							} else {
```
marcusburghardt, line 423: This new code seem pretty similar to the previous new code. Maybe for the future we should consider refactoring to remove code duplication.

marcusburghardt (commented):
I have tested in complytime-demos using a scop-config and the waived property was reflected in the assessment-plan. We can better test the report once https://github.com/oscal-compass/compliance-to-policy-go/pull/229 is merged and incorporated in complyctl.

internal/complytime/scope.go
qduanmu: Yes, one is for activity and another is for the activity step, I extracted the checkWaive part into a separate function, and no changes on other parts.

marcusburghardt (approved):
LGTM. I am looking forward to see the report with this feature in use. :)

internal/complytime/scope.go
marcusburghardt: For reference: https://github.com/complytime/complyctl/pull/269

PR #287, opened 2025-09-23
chore: [CPLYTM-1013] Ensure user workspace exists

sonupreetam:
## Summary
- Ensure user workspace exists before generate and scan
- Add PreRunE to verify user workspace via complytime.EnsureUserWorkspace
- Fails early with a clear error if missing
- Check for valid permissions

## Related Issues
- Refs CPLYTM-1013

## Changes
```go
// filepath:cli/generate.go
PreRunE: func(cmd *cobra.Command, args []string) error {
    // Ensure user workspace exists before proceeding
    if err := complytime.EnsureUserWorkspace(generateOpts.complyTimeOpts.UserWorkspace); err != nil {
        return err
    }
    return nil
}
// filepath:cli/scan.go
PreRunE: func(cmd *cobra.Command, args []string) error {
    // Ensure user workspace exists before proceeding
    if err := complytime.EnsureUserWorkspace(scanOpts.complyTimeOpts.UserWorkspace); err != nil {
        return err
    }
    return nil
},

// filepath:internal/plan.go
func EnsureUserWorkspace(userWorkspace string) error {
    if err := os.MkdirAll(userWorkspace, 0700); err != nil {
        return fmt.Errorf("failed to create user workspace directory %s: %w", userWorkspace, err)
    }
    return nil
}
// filepath:internal/plan_test.go
func TestEnsureUserWorkspace(t *testing.T) {
    tmpDir := filepath.Join(os.TempDir(), "test_workspace")
    defer os.RemoveAll(tmpDir) // Clean up after test

    err := EnsureUserWorkspace(tmpDir)
    if err != nil {
        t.Fatalf("Expected no error, got %v", err)
    }

    // Check if directory was created
    info, err := os.Stat(tmpDir)
    if err != nil {
        t.Fatalf("Expected directory to exist, got error: %v", err)
    }
    if !info.IsDir() {
        t.Fatalf("Expected a directory, got something else")
    }
}
```

### DIFF
--- a/internal/complytime/configuration.go
+++ b/internal/complytime/configuration.go
@@ -139,6 +139,16 @@ func (a ApplicationDirectory) Dirs() []string {
 	}
 }
 
+// EnsureUserWorkspace creates the user workspace directory if it doesn't exist.
+// This function should be called early in command execution to ensure the workspace
+// is available before any operations that depend on it.
+func EnsureUserWorkspace(userWorkspace string) error {
+	if err := os.MkdirAll(userWorkspace, 0700); err != nil {
+		return fmt.Errorf("failed to create user workspace directory %s: %w", userWorkspace, err)
+	}
+	return nil
+}
+
 // FindComponentDefinitions locates all the OSCAL Component Definitions in the
 // given `bundles` directory that meet the defined naming scheme.
 //
--- a/internal/complytime/configuration_test.go
+++ b/internal/complytime/configuration_test.go
@@ -50,3 +50,17 @@ func TestFindComponentDefinitions(t *testing.T) {
 	require.ErrorIs(t, err, ErrNoComponentDefinitionsFound)
 
 }
+
+func TestEnsureUserWorkspace(t *testing.T) {
+	tmpDir := t.TempDir()
+	testPlanPath := filepath.Join(tmpDir, "test_workspace")
+	defer os.RemoveAll(tmpDir) // Clean up after test
+
+	err := EnsureUserWorkspace(testPlanPath)
+	require.NoError(t, err)
+
+	info, err := os.Stat(testPlanPath)
+	require.NoError(t, err)
+	require.NotNil(t, info)
+	require.True(t, info.IsDir(), "Expected a directory, got something else")
+}
--- a/internal/complytime/plan.go
+++ b/internal/complytime/plan.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
 
 	oscalTypes "github.com/defenseunicorns/go-oscal/src/types/oscal-1-1-3"
 	"github.com/oscal-compass/oscal-sdk-go/extensions"
@@ -18,12 +17,6 @@ import (
 
 // WritePlan writes an AssessmentPlan to a given path location with consistency.
 func WritePlan(plan *oscalTypes.AssessmentPlan, frameworkId string, planLocation string) error {
-	// Ensure UserWorkspace exists before writing the plan
-	userWorkspace := filepath.Dir(planLocation)
-	if err := os.MkdirAll(userWorkspace, 0700); err != nil {
-		return err
-	}
-
 	// Add the framework property needed for ComplyTime
 	if plan.Metadata.Props == nil {
 		plan.Metadata.Props = &[]oscalTypes.Property{}
@@ -34,10 +27,8 @@ func WritePlan(plan *oscalTypes.AssessmentPlan, frameworkId string, planLocation
 		Ns:    extensions.TrestleNameSpace,
 	}
 	*plan.Metadata.Props = append(*plan.Metadata.Props, frameworkProperty)
-
 	// Handle the placeholders before writing plan
 	replacePlaceholdersInPlan(plan, frameworkId)
-
 	// To ensure we can easily read the plan once written, include under
 	// OSCAL Model type to include the top-level "assessment-plan" key.
 	oscalModels := oscalTypes.OscalModels{
--- a/cmd/complyctl/cli/generate.go
+++ b/cmd/complyctl/cli/generate.go
@@ -33,6 +33,10 @@ func generateCmd(common *option.Common) *cobra.Command {
 		Short:   "Generate PVP policy from an assessment plan",
 		Example: "complyctl generate",
 		Args:    cobra.NoArgs,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			// Ensure user workspace exists before proceeding
+			return complytime.EnsureUserWorkspace(generateOpts.complyTimeOpts.UserWorkspace)
+		},
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			return runGenerate(cmd, generateOpts)
 		},
--- a/cmd/complyctl/cli/plan.go
+++ b/cmd/complyctl/cli/plan.go
@@ -61,8 +61,10 @@ func planCmd(common *option.Common) *cobra.Command {
 		Short:   "Generate a new assessment plan for a given compliance framework id.",
 		Example: planExample,
 		Args:    cobra.ExactArgs(1),
-		PreRun: func(cmd *cobra.Command, args []string) {
+		PreRunE: func(cmd *cobra.Command, args []string) error {
 			completePlan(planOpts, args)
+			// Ensure user workspace exists before proceeding
+			return complytime.EnsureUserWorkspace(planOpts.complyTimeOpts.UserWorkspace)
 		},
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			if err := validatePlan(planOpts); err != nil {
--- a/cmd/complyctl/cli/scan.go
+++ b/cmd/complyctl/cli/scan.go
@@ -41,6 +41,10 @@ func scanCmd(common *option.Common) *cobra.Command {
 		Example:      "complyctl scan",
 		SilenceUsage: true,
 		Args:         cobra.NoArgs,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			// Ensure user workspace exists before proceeding
+			return complytime.EnsureUserWorkspace(scanOpts.complyTimeOpts.UserWorkspace)
+		},
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			return runScan(cmd, scanOpts)
 		},


internal/complytime/plan.go
```diff
@@ -16,11 +16,21 @@ import (
...
+	}
+	return nil
+}
+
 // WritePlan writes an AssessmentPlan to a given path location with consistency.
 func WritePlan(plan *oscalTypes.AssessmentPlan, frameworkId string, planLocation string) error {
 	// Ensure UserWorkspace exists before writing the plan
 	userWorkspace := filepath.Dir(planLocation)
-	if err := os.MkdirAll(userWorkspace, 0700); err != nil {
+	if err := EnsureUserWorkspace(userWorkspace); err != nil {
```
AlexXuan233: I think this is not needed since you already add `EnsureUserWorkspace` for plan command `PreRunE` func.

sonupreetam: @AlexXuan233 Thank you for your review. You absolutely right. This is redundant, I will remove it.

internal/complytime/plan_test.go
```diff
@@ -69,3 +70,21 @@ func TestPlan(t *testing.T) {
 	require.NotNil(t, ap.Metadata.Props)
 	require.Contains(t, *ap.Metadata.Props, wantProp)
 }
+
+func TestEnsureUserWorkspace(t *testing.T) {
+	tmpDir := filepath.Join(os.TempDir(), "test_workspace")
+	defer os.RemoveAll(tmpDir) // Clean up after test
+
+	err := EnsureUserWorkspace(tmpDir)
+	if err != nil {
```
gvauter: Just a nit, but `require.NoError` could be used here to keep consistency with other tests.

internal/complytime/plan_test.go
```diff
@@ -69,3 +70,21 @@ func TestPlan(t *testing.T) {
...
+		t.Fatalf("Expected no error, got %v", err)
+	}
+	// Check if directory was created
+	info, err := os.Stat(tmpDir)
+	if err != nil {
```
gvauter: Here as well.

gvauter (changes requested):
Overall looks good, left a nit on handling errors in unit tests, just to keep consistency throughout the codebase.

internal/complytime/plan.go
AlexXuan233: Just curious why remove this step in `PreRunE` for `plan` command, instead of remove this. Functionally both it's ok, but i think consistent between `plan`, `generate` and `scan` should be better. Let me know if you have different thought.

internal/complytime/plan.go
```diff
@@ -16,11 +16,21 @@ import (
 	"github.com/oscal-compass/oscal-sdk-go/validation"
 )
 
+// EnsureUserWorkspace creates the user workspace directory if it doesn't exist.
+// This function should be called early in command execution to ensure the workspace
+// is available before any operations that depend on it.
+func EnsureUserWorkspace(userWorkspace string) error {
```
marcusburghardt: Since this function is being shared by multiple commands, maybe it would be better to move it to a more generic location. Maybe in `configuration.go` , closer to the `NewApplicationDirectory` is is similar but for another context. WDYT?

internal/complytime/plan.go
sonupreetam: The main idea was to centralize workspace creation by calling EnsureUserWorkspace, reducing code duplication and keeping the logic consistent across commands. However, I am open to change this.

internal/complytime/plan_test.go
```diff
@@ -69,3 +70,18 @@ func TestPlan(t *testing.T) {
...
+
+    err := EnsureUserWorkspace(testPlanPath)
+    require.NoError(t, err)
+
+    info, err := os.Stat(testPlanPath)
+    require.NoError(t, err)
+    require.NotNil(t, info)
+    require.True(t, info.IsDir(), "Expected a directory, got something else")
+}
+
```
marcusburghardt: Nit, there are two empty lines at the end. One is enough.

marcusburghardt (changes requested):
The functionality is working fine in my tests, but besides the consistency mentioned by @AlexXuan233 , I think we could also keep the function in a more generic file.

Could you also take a look in the golangci-lint failure, please? It seems legit:
https://github.com/complytime/complyctl/actions/runs/18033430406/job/51314860100?pr=287

internal/complytime/plan.go
marcusburghardt: I believe what @AlexXuan233 is suggesting is to also move this logic to `planCmd()` as it was done with `generate` and `scan` so all commands use the same approach.

internal/complytime/plan.go
sonupreetam: I agree, while checking Duanmu's PR, I noticed this. It would be good to move it.

internal/complytime/plan.go
sonupreetam: @marcusburghardt I also asked the same question to @AlexXuan233 in Slack, sorry maybe I wasn't clear.  I will add this.

AlexXuan233: @marcusburghardt Thanks, Marcus. Yes, that's what i mean.  @sonupreetam Sorry if my comment is not very clear.

internal/complytime/plan.go
```diff
@@ -20,7 +20,7 @@ import (
 func WritePlan(plan *oscalTypes.AssessmentPlan, frameworkId string, planLocation string) error {
 	// Ensure UserWorkspace exists before writing the plan
 	userWorkspace := filepath.Dir(planLocation)
-	if err := os.MkdirAll(userWorkspace, 0700); err != nil {
+	if err := EnsureUserWorkspace(userWorkspace); err != nil {
```
AlexXuan233: Currently i think this is not needed since you already add this check to https://github.com/complytime/complyctl/pull/287/files#diff-71f94799e5fe538a6f132662c32903bf4f818ee7edcf71e18710e84b48c47780R67

internal/complytime/plan.go
sonupreetam: @AlexXuan233 Thank you for your feedback :).

internal/complytime/configuration_test.go
```diff
@@ -50,3 +50,17 @@ func TestFindComponentDefinitions(t *testing.T) {
 	require.ErrorIs(t, err, ErrNoComponentDefinitionsFound)
 
 }
+
+func TestEnsureUserWorkspace(t *testing.T) {
+	tmpDir := t.TempDir()
+	testPlanPath := filepath.Join(tmpDir, "test_workspace")
+	defer os.RemoveAll(tmpDir) // Clean up after test
+
+	err := EnsureUserWorkspace(testPlanPath)
```
marcusburghardt, line 59: Certainly not blocking this PR to be merged, but if you would like to play a little more with these tests, it would be nice to also have a case where the function fails. For example, the directory permission could be set to read-only and something like this would be expected:
```
ERROR error running complyctl: failed to create user workspace directory ./complytime: mkdir ./complytime: permission denied
```
Just an idea. : )

marcusburghardt (dismissed):
LGTM. Nice work. I have a comment regarding an additional test scenario to test a function failure situation, but only if you would like to play more with them. It is not critical in this context.

AlexXuan233 (dismissed):
LGTM

gvauter (dismissed):
LGTM

internal/complytime/configuration.go
```diff
@@ -139,6 +139,27 @@ func (a ApplicationDirectory) Dirs() []string {
 	}
 }
 
+// EnsureUserWorkspace creates the user workspace directory if it doesn't exist.
+// This function should be called early in command execution to ensure the workspace
+// is available before any operations that depend on it.
+func EnsureUserWorkspace(userWorkspace string) error {
+	if userWorkspace == "" {
```
marcusburghardt: This test is not necessary. There is already a default value for the workspace option (./complytime) in case it is not informed. In case the user explicitly use the -w option with an empty value (e.g.: `complyctl plan -w "" cusp_fedora`) it will use the current directory. So, using an empty value is not an error but a valid scenario.

internal/complytime/configuration.go
```diff
@@ -139,6 +139,27 @@ func (a ApplicationDirectory) Dirs() []string {
...
+		return fmt.Errorf("workspace path is empty")
+	}
+	if err := os.MkdirAll(userWorkspace, 0700); err != nil {
+		return fmt.Errorf("failed to create user workspace directory %q: %w", userWorkspace, err)
+	}
+	// Writable probe (captures read-only perms or read-only filesystem).
```
marcusburghardt: This code is not necessary. It increases the complexity unnecessarily. The function is focused in creating a directory, which is ensured by line 149. If the directory is new, it will be created with the proper permissions so we can infer that other functions will succeed writing new files on it. In case the directory already exists, `MkdirAll` does nothing and returns nil, which is fine. Other functions that explicitly try to write in this directory later already report error if they can't write. So, we don't need to create a file for testing this here in runtime.

internal/complytime/configuration_test.go
```diff
@@ -50,3 +50,22 @@ func TestFindComponentDefinitions(t *testing.T) {
...
+		testPlanPath := filepath.Join(tmpDir, "test_workspace")
+		err := EnsureUserWorkspace(testPlanPath)
+		require.NoError(t, err)
+		info, err := os.Stat(testPlanPath)
+		require.NoError(t, err)
+		require.True(t, info.IsDir(), "Expected a directory, got something else")
+		// Check that directory was created with correct permissions (0700)
+		require.Equal(t, os.FileMode(0700), info.Mode().Perm(), "Expected 0700 permissions")
+	})
+	t.Run("empty path fails", func(t *testing.T) {
```
marcusburghardt: This test is not necessary, as commented in the function before.

marcusburghardt (changes requested):
Hi @sonupreetam , we should stick to the previous function that was already pretty good and approved. No changes were necessary on it. The optional change in previous review was exclusively about the test scenario, to also include a failure situation when testing the new function. This optional test was not included in the last change but something else that changed the initial behavior of the function. I am fine to approve the PR as it was before and we can collaborate on the optional test in a separate PR. WDYT?

internal/complytime/configuration_test.go
sonupreetam: @marcusburghardt Maybe you must have meant to have just few lines at the end of my existing TestEnsureUserWorkspace function. Sorry didn't understand that. Thank you for your review!

marcusburghardt (approved):
LGTM. Thanks for the work here and the updates along the process @sonupreetam . I am looking forward to see your next PRs. : )

marcusburghardt:
@gvauter and @sonupreetam I just rebased it to unblock merging.

gvauter (approved):
LGTM

PR #306, opened 2025-10-01
test: expand test scenario for EnsureUserWorkspace function

marcusburghardt:
## Summary

Expanded the test to also include a situation where the function is expected to fail.
It depends on https://github.com/complytime/complyctl/pull/287

## Related Issues

It is good to test both expected and unexpected situations in our unit tests to ensure the function is properly treating these cases and other changes does not unintentionally impact this behavior.

## Review Hints

There are two commits with more context on their descriptions.
`make test-unit` should be enough for technical tests.

### DIFF
--- a/internal/complytime/configuration_test.go
+++ b/internal/complytime/configuration_test.go
@@ -54,7 +54,6 @@ func TestFindComponentDefinitions(t *testing.T) {
 func TestEnsureUserWorkspace(t *testing.T) {
 	tmpDir := t.TempDir()
 	testPlanPath := filepath.Join(tmpDir, "test_workspace")
-	defer os.RemoveAll(tmpDir) // Clean up after test
 
 	err := EnsureUserWorkspace(testPlanPath)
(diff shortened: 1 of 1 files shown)


marcusburghardt:
There was was a lint issue complaining about the line `defer os.Chmod(tmpDir, 0755)` :

```
  Error: internal/complytime/configuration_test.go:73:16: Error return value of `os.Chmod` is not checked (errcheck)
  	defer os.Chmod(tmpDir, 0755) // restore permissions so TempDir cleanup works
  	              ^
  1 issues:
  * errcheck: 1
  
  Error: issues found
```
After an investigation, I could confirm the `t.TempDir()` also deals with this situation so this line is not necessary.

sonupreetam (approved):
LGTM

sonupreetam (dismissed):
LGTM

sonupreetam (approved):
LGTM! All test passed

gvauter (approved):
LGTM

PR #314, opened 2025-10-15
chore: [CPLYTM-1102] [CPLYTM-1103] add vulnerability check

huiwangredhat:
## Summary
Add workflow for vulnerability check via [osv-scanner](https://github.com/google/osv-scanner/), which is recommended by PSIRT.
The workflow steps:
1. Install osv-scanner
2. Scan the source of `complyctl` and show the report
3. Get the `fixed version` of the CVEs
4. Upgrade the `Go version` in the gomod file.
5. Create a PR for the upgrade if there's no existing PR
  The title and the comments of the PR are similar in the example [PR](https://github.com/auto-manage-org/complyctl/pull/22
), which is created by the [run](https://github.com/auto-manage-org/complyctl/actions/runs/18517250489/job/52770087855)

## Related Issues

- _Closes #[CPLYTM-1061](https://issues.redhat.com/browse/CPLYTM-1061)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
         with:
-          go-version: '1.24.6'
+          go-version-file: 'go.mod'
 
       - name: Install cosign for signatures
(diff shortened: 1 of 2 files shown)


marcusburghardt (approved):
This is a nice work @huiwangredhat . Not only finding the issues but also creating a PR. Thanks for this.

gvauter (approved):
LGTM

sonupreetam (approved):
LGTM

PR #321, opened 2025-10-20
chore: add script to update repo workflows

qduanmu:
## Summary
This is a script to configure common workflows in a specified repo.

## Related Issues
- Closes CPLYTM-1104

## Review Hints
Need to set some env vars including GITHUB_PAT and GITHUB_USERNAME for using the github APIs or do git operations.
```
$ python scripts/setup-github-workflows.py https://github.com/{owner}/{repo}
```

### DIFF
--- a/scripts/setup-github-workflows.py
+++ b/scripts/setup-github-workflows.py
@@ -17,6 +17,17 @@ from urllib.parse import urljoin
 
 GITHUB_API = "https://api.github.com"
 GITHUB_USERNAME = os.getenv('USERNAME')
+# GitHub PAT is needed to run this script, and GitHub recommends that you use a
+# fine-grained personal access token instead of a personal access token (classic)
+# whenever possible.
+# Ref: https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?apiVersion=2022-11-28
+# The mininal required permissions for a find-grained token in this script include:
+# - Actions: read-only
+# - Administration: read and write
+# - Contents: read and write
+# - Metadata: read-only
+# - Pull requests: read and write
+# - Workflows: read and write
 GITHUB_TOKEN = os.getenv('GITHUB_PAT')
 
 # The workflow is identified by the workflow filename or ID
@@ -86,7 +97,7 @@ def push_to_origin(repo_path, workflow, owner, repo_name):
         raise RepoException(f"Git checkout failed: {e}") from e
     repo.git.add(os.path.join(WORKFLOWS_PATH, workflow + '.yml'))
 
-    commit_msg = f"chore: auto add {workflow} workflow file"
+    commit_msg = f"ci: auto add {workflow} workflow file"
     repo.index.commit(commit_msg)
     remote = repo.remote()
     remote.set_url(f"https://{GITHUB_USERNAME}:{GITHUB_TOKEN}@github.com/{owner}/{repo_name}.git")
@@ -102,7 +113,8 @@ def create_pull_request(workflow, owner, repo, workflow_branch):
                  f'-H "Authorization: Bearer {GITHUB_TOKEN}" ' \
                  '-H "X-GitHub-Api-Version: 2022-11-28"'
     data = {
-            "title": f"chore: auto workflow update for {workflow}",
+            "title": f"ci: auto workflow update for {workflow}",
+            "body": f"The update is for github {workflow} workflow.",
             "base": "main",
             "head": f"{owner}:{workflow_branch}"
     }
--- a/scripts/setup-github-workflows.py
+++ b/scripts/setup-github-workflows.py
@@ -3,15 +3,14 @@
 # Copyright Red Hat, Inc.
 
 import argparse
-import git
 import json
 import os
 import shutil
 import subprocess
+import tempfile
 
 from git import GitCommandError
 from git.repo import Repo
-from git.util import Actor
 from pathlib import Path
 from urllib.parse import urljoin
 
@@ -23,7 +22,6 @@ GITHUB_TOKEN = os.getenv('GITHUB_PAT')
 # The workflow is identified by the workflow filename or ID
 WORKFLOWS_PATH = '.github/workflows'
 COMMON_WORKFLOWS = ['codeql', 'scorecard', 'sonarcloud']
-DEST_PROJECT_ROOT = '/tmp'
 
 
 def parse_args():
@@ -78,7 +76,7 @@ def copy_workflow_file(workflow_filename, dest_project_path):
 
 
 def push_to_origin(repo_path, workflow, owner, repo_name):
-    workflow_branch = "workflow-" + workflow
+    workflow_branch = "auto-workflow-" + workflow
     repo = Repo(repo_path)
 
     print("Start to push updates to remote")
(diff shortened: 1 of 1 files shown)


scripts/setup-github-workflows.py
```diff
@@ -0,0 +1,171 @@
+#!/usr/bin/env python
+# SPDX-License-Identifier: Apache-2.0
+# Copyright Red Hat, Inc.
+
+import argparse
+import git
```
marcusburghardt: Could this import be removed since more specific imports are done below?

scripts/setup-github-workflows.py
```diff
@@ -0,0 +1,171 @@
...
+    print("Done")
+
+
+def check_workflows(repo_url):
+    owner, repo = extract_owner_repo(repo_url)
+    # Get all workflows and their states
+    cmd_header = '-H "Accept: application/vnd.github+json" ' \
+                 '-H "Authorization: Bearer {}" ' \
+                 '-H "X-GitHub-Api-Version: 2022-11-28"'.format(GITHUB_TOKEN)
+    workflows_path = '/repos/{}/{}/actions/workflows'.format(owner, repo)
```
marcusburghardt: We could also use the newer (`f'...'`) syntax for joining strings, as it was used in the previous function so the script is consistent.

scripts/setup-github-workflows.py
```diff
@@ -0,0 +1,171 @@
...
+    print("Start to push updates to remote")
+    try:
+        repo.git.checkout("-b", workflow_branch)
+    except GitCommandError as e:
+        raise RepoException(f"Git checkout failed: {e}") from e
+    repo.git.add(os.path.join(WORKFLOWS_PATH, workflow + '.yml'))
+
+    commit_msg = "chore: add {} workflow file".format(workflow)
+    committer = Actor(name=GITHUB_USERNAME, email="noreply@redhat.com")
+    commit = repo.index.commit(commit_msg, committer=committer)
```
marcusburghardt: `commit` is not used.

marcusburghardt (changes requested):
There are some minor catches. I also wonder if would be valid to automate a PR creation when a workflow needs to be created. WDYT?

scripts/setup-github-workflows.py
```diff
@@ -0,0 +1,191 @@
...
+
+from git import GitCommandError
+from git.repo import Repo
+from pathlib import Path
+from urllib.parse import urljoin
+
+
+GITHUB_API = "https://api.github.com"
+GITHUB_USERNAME = os.getenv('USERNAME')
+GITHUB_TOKEN = os.getenv('GITHUB_PAT')
```
marcusburghardt, line 31: I had to play a bit with permissions in a `fine-grained token`. For my tests, I needed these permissions:
- Actions: read-only
- Contents: read and write
- Metadata: read-only
- Pull requests: read and write
- Workflows: read and write

Maybe we could include a comment recommending `fine-grained tokens` and the minimal required permissions. WDYT?

scripts/setup-github-workflows.py
```diff
@@ -0,0 +1,191 @@
...
+    return workflow_branch
+
+
+def create_pull_request(workflow, owner, repo, workflow_branch):
+    pulls_path = f'/repos/{owner}/{repo}/pulls'
+    pulls_api = urljoin(GITHUB_API, pulls_path)
+    cmd_header = '-H "Accept: application/vnd.github+json" ' \
+                 f'-H "Authorization: Bearer {GITHUB_TOKEN}" ' \
+                 '-H "X-GitHub-Api-Version: 2022-11-28"'
+    data = {
```
marcusburghardt, line 115: Would it be possible to create a simple description, to avoid "_No description provided._" in PRs?

scripts/setup-github-workflows.py
```diff
@@ -0,0 +1,191 @@
...
+            "title": f"chore: auto workflow update for {workflow}",
```
marcusburghardt:
```suggestion
            "title": f"ci: auto workflow update for {workflow}",
```

scripts/setup-github-workflows.py
```diff
@@ -0,0 +1,191 @@
...
+    repo = Repo(repo_path)
+
+    print("Start to push updates to remote")
+    try:
+        repo.git.checkout("-b", workflow_branch)
+    except GitCommandError as e:
+        raise RepoException(f"Git checkout failed: {e}") from e
+    repo.git.add(os.path.join(WORKFLOWS_PATH, workflow + '.yml'))
+
+    commit_msg = f"chore: auto add {workflow} workflow file"
```
marcusburghardt:
```suggestion
    commit_msg = f"ci: auto add {workflow} workflow file"
```

marcusburghardt (commented):
It is working fine @qduanmu . Tested with success in a personal repository. I have only some comments about the token and minor details in the PR description and title.

marcusburghardt (dismissed):
LGTM. Thanks @qduanmu

scripts/setup-github-workflows.py
huiwangredhat: @qduanmu @marcusburghardt The `codeql` in the `COMMEN_WORKFLOWS`. But `codeql`  needs to update the settings(`code-scanning/default-setup` in line [163](https://github.com/complytime/complyctl/pull/321/files#diff-7f50d028b49451373c791ca71e05a9d9a1df1499014860e353af6548e3ae0101R163) which need the `administration` [permission](https://docs.github.com/en/rest/code-scanning/code-scanning?apiVersion=2022-11-28#update-a-code-scanning-default-setup-configuration).

huiwangredhat:
@qduanmu @marcusburghardt Thanks for your effort about re-using workflows. I have a draft idea: create a GitHub action in a new/this repo to run the python script `scripts/setup-github-workflows.py`.  In the Github action, we could use the org GitHub App token that has the temporary admin permission. We could trigger the workflow manually with the input reponame when we created a new repo under the org. WDYT? Maybe we need to update the python script a little bit and create a workflow in this repo (or create a new repo to keep the script, the github action and the reusable workflows). We could do the improvement in the next sprint.

qduanmu:
> @qduanmu @marcusburghardt Thanks for your effort about re-using workflows. I have a draft idea: create a GitHub action in a new/this repo to run the python script `scripts/setup-github-workflows.py`. In the Github action, we could use the org GitHub App token that has the temporary admin permission. We could trigger the workflow manually with the input reponame when we created a new repo under the org. WDYT? Maybe we need to update the python script a little bit and create a workflow in this repo (or create a new repo to keep the script, the github action and the reusable workflows). We could do the improvement in the next sprint.

Yes, that's what we discussed in last daily standup, a new task will be created to track the following updates, more details could be found in comment in CPLYTM-953.

scripts/setup-github-workflows.py
qduanmu: To enable it in complytime repos, permissinos need to be granted, but personal repos could be used for testing this script.

huiwangredhat: @qduanmu, thanks. I tested it in the personal repo via the `fine-grained tokens` with the permissions above, the `Configuring CodeQL `was failed with response 403. After I granted the `administration` permission to  the `fine-grained tokens`, it works. So how about updating the permissions' description?

huiwangredhat (dismissed):
LGTM. I left a comment about the `fine-grained` permission. That's not a blocker. After the improvement next , I think that's not a problem.

scripts/setup-github-workflows.py
qduanmu: I updated the comment with needed administration permissions.

huiwangredhat (approved):
LGTM. Thanks.

marcusburghardt (approved):
LGTM. Thanks

PR #327, opened 2025-10-31
chore: fix the conflict of go version

huiwangredhat:
## Summary
This PR tries to fix the [error](https://github.com/complytime/complyctl/actions/runs/18925551386/job/54147038065) about the go running version and the upgrade version in go.mod.
And it also improved to get `FIXED_VERSION` when it found vulnerabilities.

### DIFF
--- a/.github/workflows/vuln_check_fix.yml
+++ b/.github/workflows/vuln_check_fix.yml
@@ -34,13 +34,13 @@ jobs:
           # Get the CVEs fixed version from the results.sarif
           osv-scanner  -r --format=sarif . > results.sarif || true
           if [ -f "results.sarif" ] && [ -s "results.sarif" ]; then
-            FIXED_VERSION=$(jq -r '.runs[0].tool.driver.rules[] | .help.markdown | capture("### Fixed Versions\\s*\\|.+\\n\\|.+\\n\\|\\s*[^|]+?\\s*\\|\\s*[^|]+?\\s*\\|\\s*(?<version>[^|]+?)\\s*\\|") | .version' results.sarif | tr ',' '\n' | sed 's/ //g' | sort -V | tail -n 1)
+            FIXED_VERSION=$(jq -r '.runs[0].tool.driver.rules[]| .help.markdown| capture("### Fixed Versions\\s*\\|.+\\n\\|.+\\n\\|\\s*[^|]+?\\s*\\|\\s*[^|]+?\\s*\\|\\s*(?<version>[^|]+?)\\s*\\|")| .version| gsub(" "; "") | split(",")| map(split(".") | map(tonumber))| min| map(tostring)| join(".")' results.sarif | sort -V | tail -n 1)
             echo "The FIXED_VERSION is $FIXED_VERSION."
             echo "FIXED_VERSION=$FIXED_VERSION" >> $GITHUB_ENV
             echo "${{ env.FIXED_VERSION }}"
             rm -rf results.sarif
           else
-            echo "No vulnaribility"
+            echo "No vulnerability"
           fi
         env:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
@@ -82,13 +82,12 @@ jobs:
             git checkout -b "${{ env.BRANCH_NAME }}"
           fi
           go mod edit -go=${{ env.FIXED_VERSION }}
-          go mod tidy
           if [[ -z $(git status --porcelain) ]]; then
             echo "No changes to the go.mod file. Nothing to commit."
           else
             git config user.name "complytime-ci"
             git config user.email "complytime-ci@gmail.com"
-            git add go.mod go.sum
+            git add go.mod
             git commit -m "Update the gomod version to ${{ env.FIXED_VERSION }}"
             git push origin ${{ env.BRANCH_NAME }} -f
           fi


.github/workflows/vuln_check_fix.yml
```diff
@@ -34,7 +34,7 @@ jobs:
           # Get the CVEs fixed version from the results.sarif
           osv-scanner  -r --format=sarif . > results.sarif || true
           if [ -f "results.sarif" ] && [ -s "results.sarif" ]; then
-            FIXED_VERSION=$(jq -r '.runs[0].tool.driver.rules[] | .help.markdown | capture("### Fixed Versions\\s*\\|.+\\n\\|.+\\n\\|\\s*[^|]+?\\s*\\|\\s*[^|]+?\\s*\\|\\s*(?<version>[^|]+?)\\s*\\|") | .version' results.sarif | tr ',' '\n' | sed 's/ //g' | sort -V | tail -n 1)
+            FIXED_VERSION=$(jq -r '.runs[0].tool.driver.rules[]| .help.markdown| capture("### Fixed Versions\\s*\\|.+\\n\\|.+\\n\\|\\s*[^|]+?\\s*\\|\\s*[^|]+?\\s*\\|\\s*(?<version>[^|]+?)\\s*\\|")| .version| gsub(" "; "") | split(",")| map(split(".") | map(tonumber))| min| map(tostring)| join(".")' results.sarif | sort -V | tail -n 1)
```
sonupreetam, line 37: @huiwangredhat Regarding the logic change for the version picking. The previous logic split all fixed versions, sorts with sort -V, and picks the max. And in the current the min inside jq actually chooses the lowest version per rule, then you sort and pick the highest of those lows. Does this provide consistent results and is this what logic is trying to achieve here?

.github/workflows/vuln_check_fix.yml
```diff
@@ -34,7 +34,7 @@ jobs:
...
             echo "The FIXED_VERSION is $FIXED_VERSION."
             echo "FIXED_VERSION=$FIXED_VERSION" >> $GITHUB_ENV
             echo "${{ env.FIXED_VERSION }}"
```
sonupreetam, line 40: @huiwangredhat In the same run: block where you set FIXED_VERSION, you echo ${{ env.FIXED_VERSION }}. That value isn’t available until the next step, is that right? Same with $BRANCH_NAME. Please do correct me here.

.github/workflows/vuln_check_fix.yml
```diff
@@ -82,13 +82,12 @@ jobs:
...
           fi
           go mod edit -go=${{ env.FIXED_VERSION }}
-          go mod tidy
           if [[ -z $(git status --porcelain) ]]; then
             echo "No changes to the go.mod file. Nothing to commit."
           else
             git config user.name "complytime-ci"
             git config user.email "complytime-ci@gmail.com"
-            git add go.mod go.sum
+            git add go.mod
```
sonupreetam, line 90: @huiwangredhat go.sum staging and go mod tidy is removed. Is it intended? 
Also unrelated to the PR but in the file we have cat go.mod, is it necessary?

sonupreetam (changes requested):
@huiwangredhat Just added few feedbacks and some queries to review your logic efficiently.  Minor nit in the file itself "vulnaribility" to vulnerability. And should we consider filtering out pre-releases if you don’t want rc/beta Go versions, can be a separate issue, WDYT?

.github/workflows/vuln_check_fix.yml
huiwangredhat: @sonupreetam, Thanks for asking. Yes, that's the logic I want.  It will filter the fixed versions, like this 
```
{
  "version": "1.24.8, 1.25.2"
}
{
  "version": "1.24.9, 1.25.3"
}
....
``` 

For example, the first raw means the CVE could be fixed on 1.24.8 and 1.25.2. So we choose the 1.24.8. The next raw means the second CVE could be fixed on 1.24.9 and 1.25.3. We choose 1.24.9. The reasons I prefer the min one: the min one could fixed the CVE, second; the min one is closed to the project current go version, it's more safe. Finally, we will get the max one, 1.24.9 from [1.24.8, 1.24.9]. Is that helpful?

.github/workflows/vuln_check_fix.yml
huiwangredhat: @sonupreetam After setting the FIXED_VERSION to env, it works in the same block. I tested it. You could see the step "Check vulnerability" in this [run](https://github.com/auto-manage-org/complyctl/actions/runs/19054936161/job/54423065267).

.github/workflows/vuln_check_fix.yml
huiwangredhat: @sonupreetam Yes, it's intended. Removing` go.tidy`.to fix the [error](https://github.com/complytime/complyctl/actions/runs/18925551386/job/54147038065) we met.

I think `cat go.mod` is helpful. It's not necessary. The go.mod is a little bit long that makes the log is not readable :).

huiwangredhat:
> @huiwangredhat Just added few feedbacks and some queries to review your logic efficiently. Minor nit in the file itself "vulnaribility" to vulnerability. And should we consider filtering out pre-releases if you don’t want rc/beta Go versions, can be a separate issue, WDYT?

@sonupreetam Thank you. I updated the typo.  

That's a good question. The CVE has been exposed and fixed. So I think the fix version should be the release/safe version. WDYT?

.github/workflows/vuln_check_fix.yml
sonupreetam: @huiwangredhat Thank you very much for clarifying this for me. I wasn't sure we should choose the min fixed version.

.github/workflows/vuln_check_fix.yml
sonupreetam: @huiwangredhat Thank you for clarifying this. Just learnt about this from Marcus. This is good.

.github/workflows/vuln_check_fix.yml
sonupreetam: @huiwangredhat Thank you very much! Understood.

sonupreetam (approved):
@huiwangredhat Thank you for considering my feedback and responding to it. Everything LGTM.

marcusburghardt (approved):
LGTM

PR #370, opened 2026-01-15
fix: CPLYTM-1223 integration test failure

qduanmu:
## Summary
The PR is used to fixed the failed integration testing, https://github.com/complytime/complyctl/actions/runs/20945031588/job/60186283753?pr=369

## Related Issues
- Closes CPLYTM-1223

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -62,13 +62,10 @@ jobs:
           cat << EOF > config.yml
           frameworkId: cusp_fedora
           includeControls:
-            - controlId: cusp_fedora_4-1
-              controlTitle: Account Protection
+            - controlId: cusp_fedora_3-6
+              controlTitle: Files, Permissions, and Ownership
               includeRules:
-                - accounts_umask_etc_login_defs
-              selectParameters:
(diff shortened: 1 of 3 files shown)


AlexXuan233:
@qduanmu We can not just skip the  check , this will make integration test useless. We can choose another rule which can used inside a fedora container. Choose an easy change rule from https://github.com/ComplianceAsCode/content/blob/master/controls/cusp_fedora.yml

marcusburghardt (dismissed):
LGTM. Thanks @qduanmu

AlexXuan233:
@AlexXuan233 thanks, I tested with several rules before choosing this one, but I remember there is an issue if removing the condition checking, I will submit another commit for testing.

PS: Hmm, I was planning to reply, didn't notice I was actually editing the comment, paste the origin comment:

> Hi Duanmu, do you think we should also check scan result for integration-test-fedora (original) like before, currently we do not check the scan result on integration-test-fedora (original). Just a suggestion, others LGTM

AlexXuan233 (approved):
LGTM

marcusburghardt (approved):
LGTM

PR #381, opened 2026-02-17
feat(cli)!: complyctl redesign

jpower432:
## Summary

The PR implements a `complyctl` to include plugin API and UX breaking changes.

## Related Issues
_Inform any issues relevant to this PR. For example:_

- _Closes #ISSUE_NUMBER_

## Review Hints

For `specs` look at the first commit
For code changes look at the third commit

PR #380, opened 2026-02-16
feat: CPLYTM-1285 - add AMPEL plugin for complyctl

marcusburghardt:
## Summary
Implement a complyctl plugin that translates OSCAL assessment plan rules into AMPEL policy files, scans GitHub/GitLab repositories for branch protection compliance using snappy and ampel verify, and produces per-repository result files mapped to the standardized PVPResult format.

Plugin packages:
- config: Plugin configuration with path resolution
- convert: OSCAL-to-AMPEL policy conversion with CEL expressions
- scan: snappy + ampel verify CLI orchestration
- results: Output parsing, sanitization, and PVP mapping
- targets: Target repository config loading and validation
- toolcheck: Required tool presence validation
- server: policy.Provider interface implementation

Includes specification artifacts (spec, plan, tasks, data model, research, contracts) and 69 unit tests at 86.4% coverage.

## Related Issues

- New feature so we can gradually move from consuming Ampel directly to using it as complyctl plugin where we can benefit from the standardized formats and user interface offered by complyctl.
- https://issues.redhat.com/browse/CPLYTM-1285

## Review Hints

- This PR was assisted by [SpecKit](https://github.com/github/spec-kit) and ClaudeCode Opus 4.6.
- Checking the spec files is the best start to review this PR.
- The initial code was mainly created by code agents and manually reviewed to apply improvements
- I ensured all the context in the README.md, which is the best starting point to review this PR.
- For technical tests, there also a PR in complytime-demos that was developed in parallel. I recommend to use it:
  - https://github.com/complytime/complytime-demos/pull/32

PR #395, opened 2026-03-03
feat: CPLYTM-1321 - add AMPEL plugin for complyctl

marcusburghardt:
## Summary

This PR supersedes https://github.com/complytime/complyctl/pull/380. While https://github.com/complytime/complyctl/pull/380 created the initial implementation of ampel-plugin before https://github.com/complytime/complyctl/pull/381, this PR rebased the ampel plugin on top of https://github.com/complytime/complyctl/pull/381 and did all necessary changes to make it work.

## Related Issues

- ampel-plugin to check branch protection rules using gemara + complyctl + ampel.
- CPLYTM-1285
- CPLYTM-1321

## Review Hints

If you already reviewed the https://github.com/complytime/complyctl/pull/380, I recommend to see only the last commits.
If you didn't review https://github.com/complytime/complyctl/pull/380, I recommend to review the final state of this PR since the first commits were still related to OSCAL.

You can use https://github.com/complytime/complytime-demos/pull/32 to make it easier to test it in practice.


cmd/ampel-plugin/config/config.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/complytime/complyctl/internal/complytime"
+)
+
+const (
+	// PluginDir is the subdirectory name for ampel artifacts within the workspace.
+	PluginDir = "ampel"
+	// DefaultPolicyDir is the default directory name for granular AMPEL policy source files.
+	DefaultPolicyDir = "policy"
+	// GeneratedPolicyDir is the workspace subdirectory for generated policy artifacts.
+	GeneratedPolicyDir = "policy"
```
gvauter, line 20: `DefaultPolicyDir` and `GeneratedPolicyDir` are both "policy", so `PolicyDirPath()` and `GeneratedPolicyDirPath()` resolve to the same directory. This means the generated bundle is written into the same directory that `LoadGranularPolicies` reads from. It technically works because `convert.go`  has a check to skip the exact filename.  This could be a bit fragile, perhaps we use a separate directory for generated output (e.g., GeneratedPolicyDir = "generated") so source inputs and generated outputs are cleanly separated and we don't need to ensure we skip the generated file(s)?

cmd/ampel-plugin/server/server.go
```diff
@@ -0,0 +1,227 @@
...
+	generatedDir := config.GeneratedPolicyDirPath()
+	resultsDir := config.ResultsDirPath()
+	specDir := config.SpecDirPath()
+
+	if err := scan.WriteSpecFiles(specDir); err != nil {
+		return nil, fmt.Errorf("writing spec files: %w", err)
+	}
+
+	scanCfg := scan.ScanConfig{
+		PolicyPath: generatedDir + "/" + convert.PolicyFileName,
```
gvauter: nit: perhaps use `filepath.Join(generatedDir, convert.PolicyFileName)`

cmd/ampel-plugin/scan/scan.go
```diff
@@ -0,0 +1,229 @@
...
+// and spec file. The specPath must already be resolved (see ResolveSpecPath).
+func ScanRepository(repo targets.TargetRepository, branch, specPath string, cfg ScanConfig, runner CommandRunner) (*RawScanResult, error) {
+	logger := hclog.Default()
+
+	platform, org, repoName, err := targets.ParseRepoURL(repo.URL)
+	if err != nil {
+		return nil, fmt.Errorf("parsing repository URL: %w", err)
+	}
+
+	if platform != "github" {
```
gvauter: should we add `gitlab` support here as well?

gvauter (dismissed):
Left a few small comments, overall looks good!

jpower432:
@marcusburghardt I did the local testing of this with complyctl demos.

Can you help me understand the use case around `ampel-targets.yaml`? It feels like a redundant configuration surface. Seems like something being created for the plugin and not from the AMPEL ecosystem, right?

The plugin gRPC contract already sends variable to every plugin and plugins can declare this so the configuration is validated with `complyctl doctor` before scans. Would we also want to give the user the ability to use different tokens for different repos?

```
  targets:
    - id: myrepo
      policies:
        - my-source-code-policy
      variables:
        url: https://github.com/myorg/myrepo
        branch: main
        token: ${GITHUB_TOKEN}
   ```

  The specs (e.g., builtin:github/branch-rules.yaml) determine which checks run, not what to scan so it seems like it could be an assessment plan parameter.

WDYT?

File: cmd/ampel-plugin/config/config.go (hunk 1 above)
Comment (marcusburghardt) [reply cmd/ampel-plugin/config/config.go:20]: Hi George, I remember your comment in the old PR. Sure. Let me rework this structure now after rebasing. The plan is to make the directory of granular policies flexible to the user since they are not expected to be present in the system before, like with scap-security-guide for OpenSCAP. Let me check.

File: cmd/ampel-plugin/server/server.go (hunk 2 above)
Comment (marcusburghardt) [reply cmd/ampel-plugin/server/server.go]: Perfect. Thanks for catching this.

Comment (marcusburghardt) [reply cmd/ampel-plugin/server/server.go]: This is the only case I found. Fixed.

File: cmd/ampel-plugin/scan/scan.go (hunk 3 above)
Comment (marcusburghardt) [reply cmd/ampel-plugin/scan/scan.go]: It is actually not necessary this guard since snappy would already take care of this so we keep the flexibility on it. I will remove from the plugin logic. Thanks

File: cmd/ampel-plugin/config/config.go (hunk 1 above)
Comment (marcusburghardt) [reply cmd/ampel-plugin/config/config.go:20]: George, I refactored this part. Now there is a default place within the plugin workspace to host the granular policies but I also made it flexible through variables defined in `complytime.yaml`.

File: cmd/ampel-plugin/scan/scan.go (hunk 3 above)
Comment (marcusburghardt) [reply cmd/ampel-plugin/scan/scan.go]: Removed.

**Comment by marcusburghardt** [comment]
> @marcusburghardt I did the local testing of this with complyctl demos.
> 
> Can you help me understand the use case around `ampel-targets.yaml`? It feels like a redundant configuration surface. Seems like something being created for the plugin and not from the AMPEL ecosystem, right?
> 
> The plugin gRPC contract already sends variable to every plugin and plugins can declare this so the configuration is validated with `complyctl doctor` before scans. Would we also want to give the user the ability to use different tokens for different repos?
> 
> ```
>   targets:
>     - id: myrepo
>       policies:
>         - my-source-code-policy
>       variables:
>         url: https://github.com/myorg/myrepo
>         branch: main
>         token: ${GITHUB_TOKEN}
> ```
> 
> The specs (e.g., builtin:github/branch-rules.yaml) determine which checks run, not what to scan so it seems like it could be an assessment plan parameter.
> 
> WDYT?

Hi @jpower432 , thanks for asking. It was the implementation before we discussed the configuration file. I refactored this part of the plugin to get rid of `ampel-targets.yaml` and incorporated all necessary information in `complytime.yaml` so they are centralized. I imagine this configuration file may be reviewed in the future if we have multiple plugins with multiple different variables and targets that make the file too noisy, but this is not the case now and if necessary we can adapt on-demand, so I am opting to keep things simple and centralized now, aligned with your suggestion.

**Comment by marcusburghardt** [comment]
@gvauter and @jpower432 , I rebased the branch, also squashed all old commits related to the old API (OSCAL) so we keep the plugin evolution history but without noise, also implemented changes based on your feedback. The complytime-demos PR is also updated so you can easily test the recent changes. Could you take another look, please?

File: specs/001-ampel-branch-scan/contracts/plugin-manifest-schema.md
Comment (jpower432) [inline specs/001-ampel-branch-scan/contracts/plugin-manifest-schema.md:1]: Nit: This might be leftover from the OSCAL implementation

File: internal/complytime/config.go [hunk 4]
```diff
@@ -75,12 +79,21 @@ func (p PolicyEntry) EffectiveID() string {
...
 // TargetConfig binds a scan target to one or more policies with optional variables.
 // Policies are referenced by their effective ID (explicit or derived).
 type TargetConfig struct {
-	ID        string            `yaml:"id"`
-	Policies  []string          `yaml:"policies"`
-	Variables map[string]string `yaml:"variables,omitempty"`
+	ID           string            `yaml:"id"`
+	Policies     []string          `yaml:"policies"`
+	Variables    map[string]string `yaml:"variables,omitempty"`
+	Repositories []Repository      `yaml:"repositories,omitempty"`
```
jpower432: This seems like it is coupling CLI to the AMPEL plugin implementation. Is there a way to generalize this so the a user could pass in the information AMPEL needs in the existing target variables field? 
```suggestion
```

cmd/ampel-plugin/server/server.go
```diff
@@ -0,0 +1,275 @@
...
+func New() *PluginServer {
+	return &PluginServer{}
+}
+
+// Describe returns the plugin metadata and health status.
+func (s *PluginServer) Describe(_ context.Context, _ *plugin.DescribeRequest) (*plugin.DescribeResponse, error) {
+	return &plugin.DescribeResponse{
+		Healthy:                 true,
+		Version:                 "0.1.0",
+		RequiredTargetVariables: []string{},
```
jpower432: You can require target variables here
```suggestion
		RequiredTargetVariables: []string{branches, token, url},
```

jpower432 (changes requested):
@marcusburghardt I saw that the `complyctl` config was update with AMPEL specific fields, I made some suggestions in the code to clarify my suggestion. In the `Describe` a plugin can declared required target variable that can be validated against the config.

internal/complytime/config.go
marcusburghardt: Hum, so instead of passing a group of targets grouped by platform, pass each target individually and then ensure required variables. I think we can do that. At least the URL should be required. The others are optional with a default. Let me see how it would be.

cmd/mock-oci-registry/testdata/ampel-branch-protection-catalog.yaml
```diff
@@ -0,0 +1,63 @@
+title: Branch Protection Controls
+metadata:
+    id: ampel-branch-protection
```
jpower432:
```suggestion
    id: repo-branch-protection
```

cmd/mock-oci-registry/testdata/ampel-branch-protection-catalog.yaml
```diff
@@ -0,0 +1,63 @@
...
+    author:
+        id: complytime
+        name: ComplyTime
+        type: Software-Assisted
+families:
+    - id: source-code
+      title: Source Code Protection
+      description: Controls for protecting source code repositories via branch protection rules
+controls:
+    - id: bp-1
```
jpower432: Nit
```suggestion
    - id: BP-1
```

cmd/mock-oci-registry/testdata/ampel-branch-protection-catalog.yaml
```diff
@@ -0,0 +1,63 @@
...
+      title: Require Pull Request Reviews
+      objective: Protected branches must require pull request reviews before merging
+      family: source-code
+      assessment-requirements:
+          - id: SC-CODE-01.01-ar
+            text: Protected branches must require pull request reviews
```
jpower432: This is usually different than the objective. The objective is more what you are trying to achieve (or avoid) by putting this in place.
```suggestion
            text: Direct pushes to protected branches MUST be blocked
```

cmd/mock-oci-registry/testdata/ampel-branch-protection-catalog.yaml
```diff
@@ -0,0 +1,63 @@
...
+            applicability:
+                - GitHub repositories
+                - GitLab repositories
+    - id: bp-2
+      title: Require Minimum Approvals
+      objective: Pull requests to protected branches must have a minimum number of approvals
```
jpower432, line 26: This is more about code review right? The OSPSB has a good control around this - https://github.com/ossf/security-baseline/blob/bf7525fda3a70436ce58d16f67437ee794db6c51/baseline/OSPS-QA.yaml#L520. Maybe we can just import.

cmd/mock-oci-registry/testdata/ampel-branch-protection-catalog.yaml
```diff
@@ -0,0 +1,63 @@
...
+        id: complytime
+        name: ComplyTime
+        type: Software-Assisted
+families:
+    - id: source-code
+      title: Source Code Protection
+      description: Controls for protecting source code repositories via branch protection rules
+controls:
+    - id: bp-1
+      title: Require Pull Request Reviews
```
jpower432, line 15:
```suggestion
      title: Require Pull Requests
```

cmd/mock-oci-registry/testdata/ampel-branch-protection-catalog.yaml
```diff
@@ -0,0 +1,63 @@
...
+      objective: Protected branches must require pull request reviews before merging
```
jpower432:
```suggestion
      objective: Ensure changes to protected branches go through a pull request process
```

specs/001-ampel-branch-scan/contracts/plugin-manifest-schema.md
marcusburghardt: Removed

cmd/mock-oci-registry/testdata/ampel-branch-protection-catalog.yaml
```diff
@@ -0,0 +1,63 @@
...
+      family: source-code
+      assessment-requirements:
+          - id: SC-CODE-01.01-ar
```
jpower432: This is more common
```suggestion
          - id: BP-1.01
```

internal/complytime/config.go
marcusburghardt: Just implemented the changes in the last commit. In parallel updated the complytime-demos PR. You will notice I also included a `platform` variable as optional in case the target is a hosted instance and the platform is not easily determined by the `url`. The changes actually reduced the code. : )

cmd/mock-oci-registry/testdata/ampel-branch-protection-catalog.yaml
```diff
@@ -0,0 +1,63 @@
...
+                - GitHub repositories
+                - GitLab repositories
+    - id: bp-2
+      title: Require Minimum Approvals
+      objective: Pull requests to protected branches must have a minimum number of approvals
+      family: source-code
+      assessment-requirements:
+          - id: SC-CODE-02.01-ar
+            text: Pull requests must require a minimum number of approvals
+            applicability:
```
jpower432, line 31: You can include information about the applicability in `applicability-categories` in the metadata. These typically apply regardless of where the source code if hosted so should the applicability be more geared toward the sensitivity of the code or the project maturity?

cmd/ampel-plugin/server/server.go
marcusburghardt: Made only `specs` and `url` as required since the others can have reasonable default values and fallback tests.

cmd/mock-oci-registry/testdata/ampel-branch-protection-catalog.yaml
marcusburghardt: We agreed to not change much the policy in this PR.

cmd/mock-oci-registry/testdata/ampel-branch-protection-catalog.yaml
marcusburghardt: We discussed and agreed to review this in a separate PR specific for policies.

cmd/ampel-plugin/README.md
```diff
@@ -0,0 +1,250 @@
...
+### Target Configuration
+
+Repositories to scan are defined directly in `complytime.yaml` under the `repositories` field of each target entry. This eliminates the need for a separate targets file.
+
+```yaml
+targets:
+  - id: github-repos
+    policies:
+      - branch-protection
+    repositories:
```
jpower432: Nit: this is out of date

jpower432 (dismissed):
Great work @marcusburghardt. Some of the spec documentation is out of date (might be a simple fix with `speckit analyze`, but this is a non-blocking comment.

cmd/ampel-plugin/README.md
marcusburghardt: @jpower432 and @qduanmu , my last commit was only about documentation alignment. No code was touched. The plugin initially started before the complyctl redesign and was refactored to work with the new redesign. The code is fine but the documentation was not properly updated. After a few rounds of `speckit.analyse` and some reviews, it is now accurate with the current state.

jpower432 (approved):
LGTM

cmd/ampel-plugin/scan/scan.go
```diff
@@ -0,0 +1,268 @@
...
+		if !strings.HasPrefix(e, prefix) {
+			filtered = append(filtered, e)
+		}
+	}
+	filtered = append(filtered, tokenVar+"="+repo.AccessToken)
+	return filtered
+}
+
+// WriteSpecFiles writes the embedded spec files to the given directory.
+func WriteSpecFiles(specDir string) error {
```
qduanmu, line 88: Since gitlab is also supported in snappy now, this func needs to be updated to align with changes in gitlab integration.

cmd/ampel-plugin/scan/scan.go
```diff
@@ -0,0 +1,268 @@
...
+	ext := filepath.Ext(base)
+	if ext != "" {
+		base = base[:len(base)-len(ext)]
+	}
+	return base
+}
+
+// constructSnappyCommand builds the snappy snap CLI arguments for collecting
+// branch protection data from a repository using a spec file.
+func constructSnappyCommand(org, repo, branch, specPath string) []string {
```
qduanmu, line 138: The current default host is 'github.com', with the gitlab integration, different vars are used to construct the command(HOST, GROUP, PROJECT, BRANCH),  platform could be auto detected from the specpath or spec endpoint if not specified though.

cmd/ampel-plugin/server/server.go
```diff
@@ -0,0 +1,309 @@
...
+			AccessToken: accessToken,
+			Platform:    platform,
+		}
+
+		for _, branch := range branches {
+			for _, specRef := range specs {
+				specPath := scan.ResolveSpecPath(specRef, scanCfg.SpecDir)
+				logger.Info("scanning repository", "url", repoURL, "branch", branch, "spec", specRef)
+
+				rawResult, err := scan.ScanRepository(repo, branch, specPath, scanCfg, ScanRunner)
```
qduanmu, line 195: Related to comments in scan.go, if no specified, the host could be extracted from URL in repo target.

qduanmu (approved):
I left some comments with related to using the plugin with gitlab, since we pinned the snappy&ampel version(due to the go version bump in snappy) in complytime-demos, we could update this together with update in complytime-demos. Thanks for the great work, LGTM.

marcusburghardt:
> I left some comments with related to using the plugin with gitlab, since we pinned the snappy&ampel version(due to the go version bump in snappy) in complytime-demos, we could update this together with update in complytime-demos. Thanks for the great work, LGTM.

Excellent Duanmu. I propose we merge it now and create a follow-up PR adapting the whatever is necessary for GitLab since new snappy was just released with GitLab support. Let me create a task for this.

marcusburghardt:
> > I left some comments with related to using the plugin with gitlab, since we pinned the snappy&ampel version(due to the go version bump in snappy) in complytime-demos, we could update this together with update in complytime-demos. Thanks for the great work, LGTM.
> 
> Excellent Duanmu. I propose we merge it now and create a follow-up PR adapting the whatever is necessary for GitLab since new snappy was just released with GitLab support. Let me create a task for this.

https://issues.redhat.com/browse/CPLYTM-1323

PR #420, opened 2026-03-16
fix: ensures missing files are downloaded on get command execution

### LINKED ISSUE #417: complyctl get command skips re-download when cached policy directory is deleted
**Describe the bug**

`complyctl get` does not re-download a policy when its cached directory is deleted from disk. The command reports "done" for the policy but the files are not restored. This is because `SyncPolicy()` checks the digest in `state.json` but does not verify that the actual policy files exist on disk.

**To Reproduce**

1. Run `complyctl get` to sync all policies
2. Verify policies are cached: `ls ~/.complytime/policies/policies/`
3. Delete one cached policy directory: `rm -rf ~/.complytime/policies/policies/cis-fedora-l1-server`
4. Run `complyctl get` again
5. Observe output says "done" for all policies
6. Check disk: `ls ~/.complytime/policies/policies/` — the deleted policy is still missing

**Expected behavior**

`complyctl get` should detect that the policy store directory is missing on disk and re-download it from the registry, even if `state.json` still has a matching digest. The `PolicyStoreExists()` method in `internal/cache/cache.go` (line 44) already checks for the `oci-layout` marker file and could be used for this purpose.

**Screenshots**

N/A

**Additional context**

The root cause is in `internal/cache/sync.go` line 46:
```
localState, exists := s.state.GetPolicyState(policyID)
if exists && localState.Digest == remoteDigest {
    return nil
}
```

This only checks the state metadata, not whether the files exist. A suggested fix:

```
if exists && localState.Digest == remoteDigest && s.cache.PolicyStoreExists(policyID) {    
  return nil
}

As a follow-up enhancement, the output could also distinguish between "up to date" (skipped, files present) and "downloaded" (actually fetched) to give the user clearer feedback on what happened during sync.

gvauter:
## Summary

`complyctl get` silently skips re-downloading a policy when its local store directory has been deleted. The state file still records the digest as current, so the skip condition passes even though the actual content is gone.

This adds a `PolicyStoreExists` check to the skip condition so `SyncPolicy` verifies the policy store directory is present on disk before deciding to skip.

## Related Issues

Fixes https://github.com/complytime/complyctl/issues/417

## Review Hints 

- `internal/cache/sync.go`: Add `s.cache.PolicyStoreExists(policyID)` to the sync skip condition
- `internal/cache/sync_test.go`: Add `TestSync_RedownloadAfterDeletion`

> assisted by Cursor AI


marcusburghardt (approved):
LGTM. Thanks for fixing @gvauter

jpower432 (approved):
LGTM

PR #428, opened 2026-03-19
chore: align to update for gitlab nested groups in snappy

qduanmu:
## Summary
This PR aligns to [update](https://github.com/carabiner-dev/snappy/pull/43) in snappy for gitlab nested groups.


AlexXuan233 (approved):
LGTM

marcusburghardt (approved):
@qduanmu the CRAPLoad regression in `target` package seems legit. But it is fine to address this in another PR since we are just implementing crapload checks and still creating the baseline. This should also unblock other initiatives to integrate GitLab. Thanks

PR #431, opened 2026-03-23
docs: adds initial spec and plan for export functionality

gvauter:
## Summary
- Adds feature specification and implementation plan for a new `complyctl export` command that ships scan results to a ComplyBeacon OTEL collector
- Proposes extending the plugin gRPC contract with an `Export` RPC so scanning providers can emit evidence directly via ProofWatch

The export command orchestrates evidence delivery without touching evidence data itself. Plugins that opt in (via a new `supports_export` capability on `Describe`) use ProofWatch to emit `GemaraEvidence` as OTLP log records directly to the collector. Enrichment is handled downstream by TruthBeam in the Beacon collector pipeline.

Key decisions documented in the spec:

- **Standalone command** — `complyctl export --policy-id <ID>` runs after `scan`, keeping the two as composable steps
- **Plugins export directly** — complyctl passes collector config via gRPC; plugins own OTLP emission
- **Zero new deps in complyctl core** — ProofWatch/OTEL SDK only in plugins that implement Export
- **Collector config in `complytime.yaml`** — new optional `collector:` section with endpoint and auth

The implementation plan breaks work into four areas: proto contract changes, plugin SDK extension, CLI command, and config/doctor additions.

## Files

- `specs/003-complybeacon-export/spec.md` — feature specification with clarifications, user stories, acceptance scenarios, functional requirements, and success criteria
- `specs/003-complybeacon-export/plan.md` — implementation plan with project structure, proto schema, config schema, data flow, and test architecture

## Review Hints

This PR is spec and plan only — no code changes. Looking for feedback on:

1. The decision to have plugins export directly vs. complyctl exporting centrally
2. The `Export` RPC contract and `CollectorConfig` shape
3. Whether the `supports_export` capability discovery approach is sufficient
4. Any missing edge cases or requirements


specs/003-complybeacon-export/spec.md
```diff
@@ -0,0 +1,455 @@
...
+
+### Key Entities
+
+- **Beacon Collector**: A custom OpenTelemetry Collector
+  distribution (`otelcol-beacon`) that receives OTLP logs,
+  enriches them via TruthBeam, and exports to storage
+  backends (Loki, S3) and reporting tools (Hyperproof).
+
+- **ProofWatch**: A Go instrumentation library
+  (`github.com/complytime/complybeacon/proofwatch`) that
```
huiwangredhat: Nit: This library link is outdated.

specs/003-complybeacon-export/spec.md
```diff
@@ -0,0 +1,455 @@
...
+   terminal summary.
+
+---
+
+### User Story 4 - Auditor Verifies Evidence Chain (Priority: P2)
+
+An auditor needs to verify that compliance evidence in the
+downstream reporting system traces back to the original
+scan. The evidence exported by complyctl via ProofWatch
+carries integrity attributes (assessment IDs, timestamps,
```
huiwangredhat, line 286: I am not sure that Assessment IDs, timestamps, and Evaluator Identity could provide a sufficient `Chain of Custody`. To satisfy a rigorous audit, we should ensure the evidence is immutable and verifiable. Do we need to include a cryptographic hash in the results?

specs/003-complybeacon-export/spec.md
```diff
@@ -0,0 +1,455 @@
...
+
+A plugin author implements the optional `Export` RPC in their
+complyctl scanning provider. The plugin receives collector
+configuration from complyctl, creates `GemaraEvidence`
+objects from its scan results, and uses ProofWatch to emit
+them as OTLP log records to the collector. The plugin
+reports back the number of records exported and any errors.
+
+**Why this priority**: At least one plugin must implement
+export for the feature to be usable. The OpenSCAP and AMPEL
```
huiwangredhat, line 249: Just a question: are both OpenSCAP and AMPEL being developed in parallel, or will AMPEL be the first plugin to support `export`?

specs/003-complybeacon-export/spec.md
```diff
@@ -0,0 +1,455 @@
...
+  → A: The `Describe` RPC response gains a capabilities
+  field. Plugins that implement the `Export` RPC declare
+  `supports_export: true`. complyctl skips plugins that do
+  not declare this capability. Plugins that declare support
+  but fail during export return an error status without
+  affecting other plugins.
+
+- Q: Where does collector endpoint and authentication
+  configuration live? → A: `complytime.yaml` top-level
+  `collector` section with `endpoint` and `auth` fields.
```
huiwangredhat, line 63: The authtoken could be generated by the client ID and client secret dynamically, so do we need to add the two input fields?

specs/003-complybeacon-export/plan.md
```diff
@@ -0,0 +1,240 @@
...
+   c. Plugin receives ExportRequest via gRPC
+   d. Plugin initializes ProofWatch with collector endpoint
+   e. Plugin reads scan results from workspace state
+   f. Plugin creates GemaraEvidence per assessment result
+   g. Plugin calls proofwatch.Log(ctx, evidence) → OTLP
+   h. Plugin returns ExportResponse{success, counts}
+6. Display terminal summary:
+   PROVIDER        EXPORTED  FAILED  STATUS
+   openscap        47        0       ✅
+   ampel           5         0       ✅
```
huiwangredhat, line 220: Does it show the friendly error message if the status is FAILED?

huiwangredhat (commented):
@gvauter Thanks for your PR. The core logic LGTM. It's a solid implementation of the proposed plan. I’ve tagged a few minor nits and my questions.

specs/003-complybeacon-export/plan.md
```diff
@@ -0,0 +1,240 @@
+# Implementation Plan: ComplyBeacon Evidence Export
+
+**Branch**: `003-complybeacon-export` | **Date**: 2026-03-10
+**Spec**: [spec.md](spec.md)
+**Input**: Feature specification from `specs/003-complybeacon-export/spec.md`
+
+## Summary
+
+Add a `complyctl export` command that orchestrates evidence
```
jpower432: Should we make this a required feature of every plugin? Seems like it would be lightweight if it were optional. Could this be an optional "output" of `complyctl scan` instead?

specs/003-complybeacon-export/spec.md
jpower432: This is a good callout @huiwangredhat. I would think the token would generated dynamically with the `client_id`, `client_secret`, and `token_endpoint`, right? Then `complyctl` does the handshake and passes the temporary access token to the plugin which they include in the request headers along with some plugin specific identifier.

huiwangredhat: @jpower432, Oh, yes, that's right. We also need the `token_endpoint`.

specs/003-complybeacon-export/plan.md
gvauter: @jpower432 I have no strong opinions on making it mandatory.  I can't think of a plugin where we wouldn't want the evidence exported.  Are you thinking we should drop the `export` command and instead have `scan` handle the export?  I guess this depends how coupled we want those workflows to be?  A few possible thoughts
- initial users may not have the collector components running, so exporting isn't something they would want to happen by default (making them use `--skip-export` or similar flag) on every scan
- a user may want to inspect the results before exporting?  e.g. dont export failures, fix the issue, run a clean scan, then export

specs/003-complybeacon-export/plan.md
gvauter: updated FR-008 to require that the plugin's error message is displayed below the summary table when failures occur

specs/003-complybeacon-export/spec.md
gvauter: Ok thank you, I've updated this in the spec.

specs/003-complybeacon-export/spec.md
gvauter: I'm not sure which would be first, but perhaps AMPEL would be easiest to start with here?

specs/003-complybeacon-export/spec.md
gvauter: Hey Sophia - this is a good question, perhaps we can table it for now and discuss more with @jpower432 to understand the chain of custody in more detail.

specs/003-complybeacon-export/plan.md
jpower432: I agree. I was thinking it would be something controls by a flag on scan. Right now we have `--format oscal, pretty, sarif`. Could we tweak that? To me, this seems like just another way to aggregate and report the information.

It could be
```
--export oscal,sarif,pretty,otel
```

gvauter: Right now the `--format` flag is purely a complyctl-side concern, so the plugins never see this value.  If we add `otel` as a value to trigger export, do we need to rethink this design? 

One suggestion could be to have the `--format otel` flag tell complyctl to run the export phase after scan completes.  So complyctl calls the `Export()` RPC method on each capable plugin, passing CollectorConfig (endpoint + resolved token).

This keeps the design so the plugin doesn't need to interpret a format value. It implements the `Export()` RPC method, and complyctl decides when to call it.

This also leave the option open to separate these two commands in the future.  For example, I could see a common pattern where a user may run `scan` several times, remediating findings along the way.  Once the final scan is "green" they can call export.  Vs. having to rerun the scan a second time just to specify `otel` as the export method.

jpower432: 100% agree. I was thinking something really similar with the suggestion; to the user `--format` just controls how the information is aggregated and/or reported. Under the under the hood, `complyctl` would need to instruct the plugins to export.

gvauter: Ok perfect, I'll update the spec accordingly.  Thank you!

huiwangredhat (approved):
LGTM. Thank you.

jpower432 (approved):
LGTM

PR #436, opened 2026-03-25
fix: set goreleaser dir for openscap-plugin separate Go module

marcusburghardt:
## Summary
The openscap-plugin has its own go.mod and must be built from its directory.
Without `dir`, goreleaser tries to resolve the package from the root module which fails.

## Related Issues

- https://github.com/complytime/complyctl/actions/runs/23552180643/job/68568952419

## Review Hints

A lot changed with the redesign and this minor update is necessary to release.


jpower432 (approved):
LGTM

gvauter (approved):
lgtm

PR #452, opened 2026-04-01
chore: update ampel granular policy BP-02 and write spec

qduanmu:
This PR updated the Ampel granular policy file BP-02 to align to changes here: https://github.com/complytime/org-infra/pull/169 

Also, with more specs involved, updated the way to write spec files to avoid hard-coded filenames.

### DIFF
--- a/cmd/ampel-plugin/scan/scan.go
+++ b/cmd/ampel-plugin/scan/scan.go
@@ -1,10 +1,11 @@
 package scan
 
 import (
-	_ "embed"
+	"embed"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/fs"
 	"net/url"
 	"os"
 	"os/exec"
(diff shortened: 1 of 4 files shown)


cmd/ampel-plugin/scan/scan.go
```diff
@@ -30,12 +30,18 @@ var githubBranchRulesSpec []byte
...
+var gitlabProjectApprovalsSpec []byte
+
 // GitHubSpecFile is the filename for the GitHub branch rules spec.
 const GitHubSpecFile = "branch-rules.yaml"
 
 // GitLabSpecFile is the filename for the GitLab branch protection spec.
 const GitLabSpecFile = "branch-protection.yaml"
 
+// GitLabApprovalsSpecFile is the filename for the GitLab project approvals spec.
+const GitLabApprovalsSpecFile = "project-approvals.yaml"
```
marcusburghardt: I know it this PR is only incrementing what was already there, but we should think about hard-coding these file names. It should not scale well.

qduanmu: Good point, updating.

marcusburghardt (approved):
LGTM

AlexXuan233 (approved):
LGTM

PR #478, opened 2026-04-16
fix: decompose high-CRAP functions to reduce complexity

gvauter:
## Summary

Decomposes 10 functions on `main` that exceed the CRAP threshold of 30, bringing all of them well below the limit. The primary strategy is aggressive extraction of focused helper functions to reduce cyclomatic complexity to ≤ 4 (which guarantees CRAP < 30 regardless of coverage), supplemented by targeted unit tests.

| Function | Before (complexity) | After (complexity) | After CRAP |
|----------|-------------------:|-------------------:|-----------:|
| `(*scanOptions).run` | 18 | 4 | 4.0 |
| `(*generateOptions).run` | 18 | 3 | 3.0 |
| `(*getOptions).run` | 8 | 2 | 2.0 |
| `(*providersOptions).run` | 8 | 4 | 5.3 |
| `(*initOptions).run` | — | — | 28.1 |
| `(*listOptions).run` | — | — | 6.1 |
| `(*Manager).RouteGenerate` | 8 | 8 | 8.2 |
| `(*Manager).RouteScan` | 6 | 6 | 6.2 |
| OpenSCAP `(*PluginServer).Generate` | 9 | 2 | 2.1 |
| OpenSCAP `(*PluginServer).Scan` | 18 | 4 | 11.5 |

Overall CRAPload: 84 → 56.

## Related Issues

- Addresses CRAP violations flagged by the `gaze crap` CI check

## Review Hints

- Best reviewed commit-by-commit in order:
  1. **`8204be5` — CLI decomposition** (`scan.go`, `generate.go`, `get.go`, `providers.go`, new `cli_test.go`): This is the largest commit. Each `run` method now delegates to small, single-purpose helpers. No behavioral changes.
  2. **`0f7242e` — OpenSCAP plugin decomposition** (`server.go`, `server_test.go`): Same pattern applied to the OpenSCAP plugin module. Separate Go module, so fully isolated.
  3. **`13c0c37` — Plugin manager test injection** (`manager.go`, `export_test.go`, `manager_test.go`): Adds a `mockPlugin` field to `LoadedPlugin` so tests can inject a mock `Plugin` without real binaries. Adds table-driven tests for `RouteGenerate` and `RouteScan`.

- All changes are pure refactoring — no behavioral changes, no new features, no API changes.

- To verify locally:
  ```bash
  go test ./cmd/complyctl/... ./internal/... ./pkg/...
  cd cmd/openscap-plugin && go test ./...

### DIFF
--- a/cmd/complyctl/cli/cli_test.go
+++ b/cmd/complyctl/cli/cli_test.go
@@ -1,0 +1,230 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package cli
+
+import (
+	"bytes"
+	"context"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/gemaraproj/go-gemara"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/complytime/complyctl/internal/complytime"
+	"github.com/complytime/complyctl/internal/policy"
+)
+
+func chdirTemp(t *testing.T) {
+	t.Helper()
+	orig, err := os.Getwd()
+	require.NoError(t, err)
+	tmp := t.TempDir()
+	require.NoError(t, os.Chdir(tmp))
+	t.Cleanup(func() { _ = os.Chdir(orig) })
+}
+
+func writeWorkspaceConfig(t *testing.T, content string) {
+	t.Helper()
+	require.NoError(t, os.WriteFile("complytime.yaml", []byte(content), 0600))
+}
+
+const minimalConfig = `policies:
+  - url: registry.example.com/policies/test-policy@v1.0
+    id: test-policy
+targets:
+  - id: local
+    policies:
+      - test-policy
+    variables:
+      profile: test
+`
+
+// --- scanOptions tests ---
+
+func TestScanOptions_Run_NoWorkspace(t *testing.T) {
+	chdirTemp(t)
+	o := &scanOptions{
+		Common:  &Common{},
+		timeout: 5 * time.Second,
+	}
+	err := o.run(context.Background())
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "failed to load workspace config")
+}
+
+func TestScanOptions_Run_NoTargets(t *testing.T) {
+	chdirTemp(t)
+	writeWorkspaceConfig(t, `policies:
+  - url: registry.example.com/policies/test@v1.0
+targets: []
+`)
+	o := &scanOptions{
+		Common:  &Common{},
+		timeout: 5 * time.Second,
+	}
+	err := o.run(context.Background())
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "no targets")
+}
+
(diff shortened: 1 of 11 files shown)


pkg/plugin/manager.go
```diff
@@ -30,8 +30,9 @@ type Manager struct {
 
 // LoadedPlugin pairs discovery metadata with a live gRPC client.
 type LoadedPlugin struct {
-	Info   PluginInfo
-	Client *Client
+	Info       PluginInfo
+	Client     *Client
+	mockPlugin Plugin // injected during tests; nil in production
```
marcusburghardt: Nit (non-blocking): This adds a test-only field to a production struct. An alternative would be to define a `PluginClient` interface at the Manager level and inject mocks through that, keeping `LoadedPlugin` free of test-only fields. This would better align with Principle II (Simplicity & Isolation) by separating test and production concerns.

Not in scope for this PR — just flagging for a potential follow-up.

cmd/complyctl/cli/scan.go
```diff
@@ -10,6 +10,7 @@ import (
 	"slices"
 	"time"
 
+	"github.com/gemaraproj/go-gemara"
```
marcusburghardt: Nit (non-blocking): This introduces a direct import of `go-gemara` in scan.go for the `gemara.EvaluationLog` type used in the extracted report helpers. Previously, this type was only accessed transitively through the `output` package. Consider whether the report-writing helpers (`writePrettyReport`, `writeSARIFReport`, `writeOSCALReport`) could accept a narrower interface or remain closer to the `output` package to avoid spreading the direct dependency.

Not in scope for this PR — just a coupling observation.

marcusburghardt (commented):
AI-assisted review — two non-blocking nits flagged inline for future consideration.

cmd/complyctl/cli/scan.go
gvauter: Thanks @marcusburghardt - I've added an additional commit to address these comments.

marcusburghardt (approved):
LGTM. We can update the baseline in a separate PR. Thanks for addressing this @gvauter

cmd/complyctl/cli/cli_test.go
```diff
@@ -0,0 +1,230 @@
...
+
+	"github.com/gemaraproj/go-gemara"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/complytime/complyctl/internal/complytime"
+	"github.com/complytime/complyctl/internal/policy"
+)
+
+func chdirTemp(t *testing.T) {
```
jpower432, line 20: Nit: I think testing.T has a function for this t.Chdir()

jpower432 (approved):
LGTM

cmd/complyctl/cli/cli_test.go
gvauter: I will fix in an upcoming PR.  TY!

PR #479, opened 2026-04-17
fix: get resolves pinned version instead of defaulting to latest tag

jpower432:
## Summary

SyncPolicy passed the bare policyID to DefinitionVersion, which unconditionally appended :latest. Registries without a latest tag failed even when the pinned version existed. DefinitionVersion now only appends :latest when the modulePath has no explicit version, and SyncPolicy constructs policyID:version when a version is specified.

## Related Issues

N/A

## Review Hints

Add a version the policy ref in the `complytime.yaml` (example: `@v1.0.0`). It currently fails and appends `:latest` to it. With this change `@` syntax should be fully supported.

### DIFF
--- a/internal/cache/sync.go
+++ b/internal/cache/sync.go
@@ -30,7 +30,12 @@ func (s *Sync) SyncPolicy(ctx context.Context, policyID, version string) error {
 		return fmt.Errorf("policy ID cannot be empty")
 	}
 
-	remoteDigest, remoteVersion, err := s.source.DefinitionVersion(ctx, policyID)
+	lookupRef := policyID
+	if version != "" && version != "latest" {
+		lookupRef = policyID + ":" + version
+	}
+
+	remoteDigest, remoteVersion, err := s.source.DefinitionVersion(ctx, lookupRef)
 	if err != nil {
 		return fmt.Errorf(
 			"policy %s: registry unreachable: %w (cached data may still be available)",
--- a/internal/doctor/doctor.go
+++ b/internal/doctor/doctor.go
@@ -55,6 +55,7 @@ type PolicyGraphResolver interface {
 // See R55: specs/001-gemara-native-workflow/spec.md
 type VersionResolver interface {
 	ResolveLatestVersion(registry, repository string) (version string, err error)
+	ResolveVersion(registry, repository, version string) (string, error)
 }
 
 const registryTimeout = 5 * time.Second
@@ -253,11 +254,23 @@ func CheckPolicyVersions(cfg *complytime.WorkspaceConfig, cacheDir string, versi
 
 		latestVersion, err := versionResolver.ResolveLatestVersion(ref.Registry, ref.Repository)
 		if err != nil {
+			if ref.Version != "" {
+				_, pinnedErr := versionResolver.ResolveVersion(ref.Registry, ref.Repository, ref.Version)
+				if pinnedErr == nil {
+					results = append(results, CheckResult{
+						Name:     fmt.Sprintf("policy/%s", eid),
+						Status:   StatusPass,
+						Message:  fmt.Sprintf("%s (pinned — latest tag unavailable for staleness check)", cachedState.Version),
+						Blocking: false,
+					})
+					continue
+				}
+			}
 			unreachable[ref.Registry] = true
 			results = append(results, CheckResult{
 				Name:     fmt.Sprintf("registry/%s", ref.Registry),
 				Status:   StatusWarn,
-				Message:  "unreachable — version check skipped",
+				Message:  fmt.Sprintf("unreachable: %v", err),
 				Blocking: false,
 			})
 			continue
--- a/internal/doctor/doctor.go
+++ b/internal/doctor/doctor.go
@@ -49,12 +49,16 @@ type PolicyGraphResolver interface {
 	ResolvePolicyGraph(policyID, version string) (*policy.DependencyGraph, error)
 }
 
-// VersionResolver queries an OCI registry for the latest version of a policy.
+// VersionResolver queries an OCI registry for policy version information.
+// It supports both latest-tag resolution for staleness checks and pinned
+// version resolution for reachability verification.
 // Satisfied by the adapter in cmd/complyctl/cli/doctor.go — defined as
 // interface for testability (Constitution II).
 // See R55: specs/001-gemara-native-workflow/spec.md
 type VersionResolver interface {
+	// ResolveLatestVersion resolves the latest tag for staleness comparison.
 	ResolveLatestVersion(registry, repository string) (version string, err error)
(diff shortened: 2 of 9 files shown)


internal/registry/client.go
```diff
@@ -53,7 +53,10 @@ func (c *Client) DefinitionVersion(ctx context.Context, modulePath string) (stri
 		return c.fetcher.DefinitionVersion(ctx, modulePath)
 	}
 
-	ref := fmt.Sprintf("%s/%s:latest", c.registryHost(), modulePath)
+	ref := fmt.Sprintf("%s/%s", c.registryHost(), modulePath)
+	if !strings.Contains(modulePath, ":") && !strings.Contains(modulePath, "@") {
+		ref += ":latest"
```
marcusburghardt: **[HIGH] CRAP regression: `DefinitionVersion` crossed CRAPload threshold (6.7 → 15.3)**

The new `strings.Contains` branching (lines 56-58) is untested — all existing tests in `client_test.go` use `NewClientWithFetcher` with a mock, which returns at the `c.fetcher != nil` guard (line 52) before reaching this logic.

Consider adding tests that use a `Client` without a fetcher (or with a nil fetcher) and exercise three paths:
1. `modulePath` without `:` or `@` → should append `:latest`
2. `modulePath` with `:` (e.g., `"policies/nist:v1.0.0"`) → should NOT append `:latest`
3. `modulePath` with `@` (e.g., `"policies/nist@sha256:abc"`) → should NOT append `:latest`

Since these hit a real registry resolve, you'd likely need to test at the ref-construction level (e.g., extract a `buildRef` helper that's independently testable) or use `fetchVersion` with a test server.

internal/cache/sync.go
```diff
@@ -30,7 +30,12 @@ func (s *Sync) SyncPolicy(ctx context.Context, policyID, version string) error {
 		return fmt.Errorf("policy ID cannot be empty")
 	}
 
-	remoteDigest, remoteVersion, err := s.source.DefinitionVersion(ctx, policyID)
+	lookupRef := policyID
+	if version != "" && version != "latest" {
+		lookupRef = policyID + ":" + version
+	}
```
marcusburghardt, line 36: **[MEDIUM] Missing test coverage for version-specific ref construction**

The new branching (`version != "" && version != "latest"` → `policyID + ":" + version`) is not exercised by existing sync tests — they all call `SyncPolicy` with `"latest"`. CRAP score rose from 11.4 → 13.9.

A test like `TestSync_CopyOnSuccess_PinnedVersion` calling `SyncPolicy(ctx, "test-policy", "v1.0.0")` and verifying that the source receives the `"test-policy:v1.0.0"` ref would cover this path.

internal/doctor/doctor_test.go
```diff
@@ -231,6 +249,76 @@ func TestCheckPolicyVersions_RegistryUnreachable(t *testing.T) {
 	if results[0].Status != StatusWarn {
 		t.Errorf("expected warn, got %s", results[0].Status)
 	}
+	if !strings.Contains(results[0].Message, "connection refused") {
```
marcusburghardt, line 252: **[INFO] Good coverage — two new test cases cover the pinned-version fallback**

`TestCheckPolicyVersions_LatestMissing_PinnedResolves` (positive) and `TestCheckPolicyVersions_LatestMissing_NoPinnedVersion` (negative) properly verify both outcomes of the new fallback logic. The existing `TestCheckPolicyVersions_RegistryUnreachable` was also updated to assert the actual error message. This is the model the other changed files should follow.

marcusburghardt (commented):
AI-assisted review: 2 test coverage gaps identified (CRAP regressions), 1 positive note on doctor tests.

marcusburghardt:
@jpower432 a rebase on this should resolve the CI failures. :)

internal/doctor/doctor.go
```diff
@@ -55,6 +55,7 @@ type PolicyGraphResolver interface {
 // See R55: specs/001-gemara-native-workflow/spec.md
 type VersionResolver interface {
 	ResolveLatestVersion(registry, repository string) (version string, err error)
+	ResolveVersion(registry, repository, version string) (string, error)
```
marcusburghardt, line 62: **[MEDIUM] Missing GoDoc on exported interface method (CS-004 / DR-001)**

`ResolveVersion` is added to the exported `VersionResolver` interface without a GoDoc comment. Convention pack CS-004 requires GoDoc on all exported identifiers. Additionally, the interface-level comment ("queries an OCI registry for the latest version") no longer fully describes the interface's scope now that it also supports pinned version resolution.

Suggested fix: add a GoDoc comment on `ResolveVersion` and update the interface comment to reflect both capabilities.

internal/doctor/doctor.go
```diff
@@ -253,11 +254,23 @@ func CheckPolicyVersions(cfg *complytime.WorkspaceConfig, cacheDir string, versi
...
+			if ref.Version != "" {
+				_, pinnedErr := versionResolver.ResolveVersion(ref.Registry, ref.Repository, ref.Version)
+				if pinnedErr == nil {
+					results = append(results, CheckResult{
+						Name:     fmt.Sprintf("policy/%s", eid),
+						Status:   StatusPass,
+						Message:  fmt.Sprintf("%s (pinned — latest tag unavailable for staleness check)", cachedState.Version),
+						Blocking: false,
+					})
+					continue
```
marcusburghardt: **[MEDIUM] CRAP regression — consider extracting fallback logic**

This fallback block (lines 256-277) adds conditional branching that increased `CheckPolicyVersions` CRAP score from 10 to 12, exceeding the regression epsilon of 0.5. Extracting this into a helper function (e.g., `resolveVersionFallback`) would reduce the function's cyclomatic complexity and resolve the CRAP Load CI gate failure.

marcusburghardt (changes requested):
PR introduces a clean fix for pinned version resolution. Code is correct, well-tested, and aligned with intent. Three MEDIUM findings noted below.

The CRAP Load CI gate failed due to 2 regressions: `SyncPolicy` (11.4 -> 13.4, +2.0) and `CheckPolicyVersions` (10 -> 12, +2.0). Both remain well under the 30 threshold. The regressions are inherent to the added branching logic. Consider extracting the fallback logic in `CheckPolicyVersions` into a helper to reduce complexity. If the team prefers to accept the regression instead, please update the CRAP baseline (`.gaze/baseline.json`) as part of this PR so the CI gate passes before merge.

Commits 1 and 2 use `Made-with: Cursor` but lack the required `Assisted-by` trailer per the constitution's Commit Trailers section. Commit 3 correctly includes both.

marcusburghardt (approved):
LGTM. Thanks @jpower432

sonupreetam (approved):
LGTM

PR #490, opened 2026-04-27
feat(registry): resolve policy definitions by tag or digest

sonupreetam:
## Summary

- Extend `DefinitionVersion` (registry client, `PolicySource`, mocks) with a `ref` argument so policies can be resolved by **tag** or **`sha256:...` manifest digest**, not only `:latest`.
- `SyncPolicy` passes the configured version, defaulting the empty case to `latest` for resolution.
- CLI `doctor` / `get` call sites pass an empty ref for existing “resolve latest” behavior.
- Tests: updated mocks, `TestClient_DefinitionVersion_ExplicitRef` for a specific tag.

## Notes

- `complytime.yaml` local OCI test edits were **not** included; the sample file remains the upstream default.

**Branch:** `sonupreetam:feat/registry-definition-version-ref` → `complytime/complyctl:main` (draft)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/cmd/complyctl/cli/doctor.go
+++ b/cmd/complyctl/cli/doctor.go
@@ -46,7 +46,7 @@ func (r *registryVersionResolver) ResolveLatestVersion(registryURL, repository s
 	client := registry.NewClient(registryURL, credFunc)
 	ctx, cancel := context.WithTimeout(context.Background(), r.timeout)
 	defer cancel()
-	_, version, err := client.DefinitionVersion(ctx, repository)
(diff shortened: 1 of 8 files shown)


sonupreetam:
Closing — superseded by #479 which covers the same pinned-version/digest resolution logic (via `buildRef()` helper) without breaking the `DefinitionVersion` signature. The only additive value here was mock/test scaffolding that is likely to be replaced, so not worth maintaining a separate PR for.

PR #510, opened 2026-05-13
feat: add errors field to ScanResponse for improved error reporting

jpower432:
## Summary

  • ScanResponse was the only RPC response without an error surface. Providers had no structured way to report "I couldn't evaluate this target" separately from "I evaluated this target and it failed." Both got flattened into the same AssessmentLog stream.
  • Adds repeated string errors to ScanResponse in the proto, ScanResult/RouteScanResult() to the manager, and wires it into the CLI so complyctl scan exits non-zero on operational errors while policy violations continue to exit 0.

  The complyctl gRPC adapter (grpcServer.Scan()) discards the response when Scan() returns a non-nil error. This is a gRPC protocol constraint. Providers that encountered partial failures
  (e.g. 3 of 5 targets scanned) had two bad options:

  1. Return (nil, error) — discard all partial results
  2. Return (response, nil) — smuggle errors into synthetic assessment entries with requirement IDs like scan-error, indistinguishable from real evaluation results

  Solution

  Two channels in ScanResponse, matching the pattern already used by GenerateResponse, DescribeResponse, and ExportResponse.

 **Exit code semantics**:

  | All targets scanned, all passed           │ 0         │
  │ All targets scanned, some policies failed │ 0         │
  │ Some targets had operational errors       │ non-zero  │

  Policy violations are findings (data). Operational errors are failures (the tool couldn't do its job). Reports are written before the error exit so partial results remain available.

## Related Issues

- _Closes #ISSUE_NUMBER_

## Review Hints

  Test plan

  - [ ] RouteScanResult returns errors from provider ScanResponse.errors
  - [ ] RouteScanResult captures RPC failures in both Errors and Assessments
  - [ ] RouteScanResult with clean scan has empty Errors
  - [ ] FormatOperationalWarnings renders distinct warnings block
  - [ ] FormatOperationalWarnings returns empty string for no errors
  - [ ] Existing scan tests pass unchanged (backward compatible)
  - [ ] make lint clean


CHANGELOG.md
```diff
@@ -1 +1,18 @@
 # Changelog
```
hbraswelrh, line 1: Was this intentionally added? I hadn't noticed it being used outside of releases.

jpower432: No. I will remove it.

pkg/provider/manager.go
```diff
@@ -190,25 +214,33 @@ func (m *Manager) RouteScan(ctx context.Context, evaluatorID string, targets []T
 			msg := m.scanErrorMessage(p.Info.ProviderID, scanErr, ctx)
 			m.logger.Error("Provider Scan failed",
 				"provider_id", p.Info.ProviderID, "error", scanErr)
-			return errorAssessments(evaluatorID, msg), nil
+			return &ScanResult{
+				Assessments: errorAssessments(evaluatorID, msg),
+				Errors:      []string{msg},
+			}, nil
```
sonupreetam, line 237: **[HIGH] Operational errors are double-counted in scan output**

When a provider Scan RPC fails, `RouteScanResult` populates both `Assessments` (via `errorAssessments()`) and `Errors` with the same message. Downstream, `processScanOutput` feeds both into different renderers:

1. `scanOut.errors` → `reportOperationalWarnings()` → stderr WARNING block
2. `scanOut.assessments` (including the synthetic error entry) → `FormatScanSummary()` → table row with ⚠️ + totals line counting `1 error`

The user sees the same failure twice in different visual forms. The summary table's `error` count is inflated by something that isn't an evaluation result.

The same duplication occurs in the broadcast path (line ~236).

The design doc (D3) states: *"They are not mixed into the assessment table rows."* — but `errorAssessments()` still injects them into the assessment stream.

Suggested fix: `RouteScanResult` should only populate `.Errors` on RPC failure and skip `errorAssessments()`. Let `RouteScan()` (the backward-compatible wrapper) call `errorAssessments()` itself so legacy callers still get the synthetic entries.

sonupreetam (changes requested):
**Verdict: REQUEST CHANGES**

Strong design with good spec artifacts and comprehensive tests. One substantive issue: operational errors from RPC failures appear in both the WARNING block and the assessment summary table (`errorAssessments` populates `ScanResult.Assessments` alongside `ScanResult.Errors`). This contradicts design doc D3 ("not mixed into the assessment table rows") and inflates the scan summary's error count.

api/plugin/plugin.pb.go
```diff
@@ -16,6 +16,7 @@ import (
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 	reflect "reflect"
 	sync "sync"
+	unsafe "unsafe"
```
trevor-vaughan: Can we prevent making this package `unsafe`?

jpower432: Yea, it is being using for string processing, but this a generated file. I will look into what is causing it.

jpower432:
@sonupreetam @trevor-vaughan PTAL

sonupreetam (approved):
LGTM

hbraswelrh (approved):
LGTM.

jpower432:
Rebased with current main

PR #514, opened 2026-05-18
fix(doctor): improve UX for version resolution errors and diagnostics

### LINKED ISSUE #503: fix: improve UX for version resolution error messages and doctor diagnostics
## Summary

During review of #479 (pinned version resolution fix), end-to-end testing of `complyctl get` and `complyctl doctor` across several version/registry configurations revealed UX issues in error messages and diagnostic output. These are pre-existing or only partially addressed by #479.

### 1. `doctor` warns about staleness on deliberately pinned versions

**Scenario**: User configures `url: registry.io/policies/nist@v1.0.0` (pinned). Registry has both `v1.0.0` and `latest`.

**Current output**:
```
⏭️ policy/nist-pinned: cached v1.0.0, available latest — run complyctl get to update
```

**Problem**: The user deliberately pinned to `v1.0.0`. Suggesting they "update" contradicts their intent and creates unnecessary noise. Running `get` would be a no-op since the pinned version is already cached.

**Expected**: Doctor should respect the pin and suppress the staleness warning, or adjust the message to acknowledge the pin (e.g., `✅ policy/nist-pinned: v1.0.0 (pinned — latest available: v1.0.0)`).

### 2. "registry unreachable" when only the `latest` tag is missing

**Scenario**: User configures `url: registry.io/policies/test` (no version). Registry is online but has no `latest` tag (only versioned tags like `v1.0.0`).

**Current output**:
```
⏭️ registry/http://registry.io: unreachable: ...OCI version resolution failed for .../policies/test:latest: not found
```

**Problem**: The registry is reachable — the tag just doesn't exist. "unreachable" is a misdiagnosis. The user has no guidance on how to fix this.

**Expected**: Something like `⏭️ policy/test: latest tag not found — pin a specific version with @v1.0.0`

### 3. Error messages have excessive nesting and internal detail leakage

**Scenario**: `complyctl get` fails for a missing version.

**Current output**:
```
Error: failed to sync policy policies/no-latest: policy policies/no-latest: registry unreachable: failed to fetch version for policies/no-latest: OCI version resolution failed for localhost:8766/policies/no-latest:latest: localhost:8766/policies/no-latest:latest: not found (cached data may still be available) (cached policies: [policies/cis-fedora-l1-workstation ...])
```

**Problem**: 5 layers of `fmt.Errorf` wrapping. The policy name appears 4 times. Internal details (full OCI ref format, unrelated cached policy names) are exposed. Users must parse through all of this to find "not found."

**Expected**: Flatten to a user-friendly message, e.g.: `Error: policy no-latest: version "latest" not found in registry localhost:8766`

### 4. `doctor` doesn't surface configured-vs-cached version mismatch

**Scenario**: User configures `@v9.9.9` (doesn't exist). A previous `get` cached `v1.0.0`.

**Current output**:
```
⏭️ policy/nist-wrong: cached v1.0.0, available latest — run complyctl get to update
```

**Problem**: Doctor compares cached version against `latest`, ignoring that the user configured `v9.9.9`. Running `get` as suggested will fail because `v9.9.9` doesn't exist. Doctor should detect the mismatch between configured version and cached version.

### 5. `doctor` panics with no `complytime.yaml`

**Scenario**: Run `complyctl doctor` in a directory with no `complytime.yaml`.

**Current output**:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x40 pc=...]

goroutine 1 [running]:
github.com/complytime/complyctl/internal/doctor.CheckCollector(0x...)
        internal/doctor/doctor.go:678 +0x20
```

**Problem**: `runDoctor` passes a nil `cfg` to `doctor.Run` when config loading fails. `CheckCollector` dereferences it without a nil check.

**Expected**: Doctor should report `❌ config: complytime.yaml not found — run complyctl init` and continue remaining checks gracefully (or exit cleanly).

## Reproduction

All scenarios were tested locally with the mock OCI registry (`cmd/mock-oci-registry`) and a custom no-latest-tag registry. Steps:

1. `make build`
2. Start mock registry: `make mock-registry-background`
3. Create `complytime.yaml` with the scenario config (see descriptions above)
4. Run `complyctl get` and `complyctl doctor`

## Suggested Approach

- Issues 1-4 can be addressed together in a single PR focused on `internal/doctor/doctor.go` and `internal/cache/sync.go` error messaging
- Issue 5 (nil panic) is a separate bug fix — a nil guard in `CheckCollector` and any other check that receives `cfg`

Found during review of #479.

sonupreetam:
## Summary

Addresses all 5 sub-issues from #503:

- **Pinned version staleness**: Doctor no longer warns "run complyctl get to update" on deliberately pinned versions. Reports `(pinned)` with optional latest-availability note.
- **Misdiagnosed "registry unreachable"**: New `ErrVersionNotFound` sentinel classifies 404 responses, enabling accurate messages like "latest tag not found — pin a specific version with @<tag>".
- **Error nesting flattened**: Removes redundant `"failed to sync policy … registry unreachable … (cached data may still be available) (cached policies: [...])"` wrapping. Errors pass through cleanly. Non-404 errors retain module path context.
- **Configured-vs-cached mismatch**: If a user pins `@v2.0.0` but cache holds `v1.0.0`, doctor reports the mismatch with guidance to run `complyctl get`.
- **Nil panic in CheckCollector**: `cfg == nil` guard prevents SIGSEGV when `complytime.yaml` is absent.

Closes #503

## Changes

| File | Purpose |
|------|---------|
| `internal/registry/client.go` | `ErrVersionNotFound` sentinel; classify 404 vs network errors; restore context wrapping for non-404 |
| `internal/cache/sync.go` | Flatten error wrapping; distinguish "not found" from "unreachable" |
| `internal/doctor/doctor.go` | Pinned-version logic, mismatch detection, `resolvePinnedFallback` improvement, nil guard, 404 poison-guard fix |
| `internal/doctor/doctor_test.go` | Tests for all 5 scenarios + regression tests (poison-guard, pinned+both 404, mixed registries, pinned network failure) |
| `cmd/complyctl/cli/get.go` | Remove redundant error re-wrapping and `suggestCachedPolicyIDs` |
| `cmd/complyctl/cli/cli_test.go` | Update assertion to match flattened error format |

## Test plan

- [x] `go test ./internal/doctor/` — all 17 `TestCheckPolicyVersions*` tests pass
- [x] `go test ./internal/cache/ ./internal/registry/ ./cmd/complyctl/cli/` — all pass
- [x] `go build ./...` — compiles cleanly
- [ ] CI unit tests
- [ ] E2E tests with mock registry (pinned, unpinned, missing-tag, mismatch scenarios)

### DIFF
--- a/internal/doctor/doctor.go
+++ b/internal/doctor/doctor.go
@@ -262,7 +262,7 @@ func CheckPolicyVersions(cfg *complytime.WorkspaceConfig, cacheDir string, versi
 		latestVersion, err := versionResolver.ResolveLatestVersion(ref.Registry, ref.Repository)
 		if err != nil {
 			result := resolvePinnedFallback(versionResolver, ref, eid, cachedState.Version, err)
-			if result.Status == StatusWarn {
+			if result.Status == StatusWarn && !errors.Is(err, registry.ErrVersionNotFound) {
 				unreachable[ref.Registry] = true
 			}
 			results = append(results, result)
--- a/internal/doctor/doctor.go
+++ b/internal/doctor/doctor.go
@@ -4,6 +4,7 @@ package doctor
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -14,6 +15,7 @@ import (
 	"github.com/complytime/complyctl/internal/cache"
 	"github.com/complytime/complyctl/internal/complytime"
 	"github.com/complytime/complyctl/internal/policy"
+	"github.com/complytime/complyctl/internal/registry"
 	"github.com/complytime/complyctl/pkg/provider"
 )
 
@@ -268,7 +270,28 @@ func CheckPolicyVersions(cfg *complytime.WorkspaceConfig, cacheDir string, versi
 		}
 
 		cachedVersion := cachedState.Version
-		if cachedVersion == latestVersion {
+
+		if ref.Version != "" {
+			if cachedVersion == ref.Version {
+				msg := fmt.Sprintf("%s (pinned)", cachedVersion)
+				if latestVersion != cachedVersion {
+					msg = fmt.Sprintf("%s (pinned — latest available: %s)", cachedVersion, latestVersion)
+				}
+				results = append(results, CheckResult{
+					Name:     fmt.Sprintf("policy/%s", eid),
+					Status:   StatusPass,
+					Message:  msg,
+					Blocking: false,
+				})
+			} else {
+				results = append(results, CheckResult{
+					Name:     fmt.Sprintf("policy/%s", eid),
+					Status:   StatusWarn,
+					Message:  fmt.Sprintf("cached %s does not match configured pin @%s — run complyctl get", cachedVersion, ref.Version),
+					Blocking: false,
+				})
+			}
+		} else if cachedVersion == latestVersion {
 			results = append(results, CheckResult{
 				Name:     fmt.Sprintf("policy/%s", eid),
 				Status:   StatusPass,
@@ -290,7 +313,7 @@ func CheckPolicyVersions(cfg *complytime.WorkspaceConfig, cacheDir string, versi
 
 // resolvePinnedFallback attempts to resolve a pinned version when the latest
 // tag is unavailable. Returns a pass result if the pinned version resolves,
-// or a warn result marking the registry as unreachable.
+// or a warn result with a user-friendly diagnosis.
 func resolvePinnedFallback(
 	resolver VersionResolver,
 	ref complytime.PolicyRef,
@@ -308,6 +331,20 @@ func resolvePinnedFallback(
 			}
 		}
 	}
+
+	if errors.Is(latestErr, registry.ErrVersionNotFound) {
+		msg := "latest tag not found — pin a specific version with @<tag>"
+		if ref.Version != "" {
+			msg = fmt.Sprintf("version %q not found in registry", ref.Version)
+		}
+		return CheckResult{
+			Name:     fmt.Sprintf("policy/%s", eid),
+			Status:   StatusWarn,
+			Message:  msg,
+			Blocking: false,
+		}
+	}
+
 	return CheckResult{
 		Name:     fmt.Sprintf("registry/%s", ref.Registry),
 		Status:   StatusWarn,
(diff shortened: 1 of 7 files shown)


internal/registry/client.go
```diff
@@ -56,7 +62,7 @@ func (c *Client) DefinitionVersion(ctx context.Context, modulePath string) (stri
 	ref := buildRef(c.registryHost(), modulePath)
 	digest, version, err := c.fetchVersion(ctx, ref)
 	if err != nil {
-		return "", "", fmt.Errorf("failed to fetch version for %s: %w", modulePath, err)
+		return "", "", err
```
jpower432: Was it intentional to strip the error context?

internal/doctor/doctor.go
```diff
@@ -308,6 +331,20 @@ func resolvePinnedFallback(
 			}
 		}
 	}
+
+	if errors.Is(latestErr, registry.ErrVersionNotFound) {
```
jpower432, line 335: This new `ErrVersionNotFound` branch returns `StatusWarn`, which causes the existing caller (line 265) to set `unreachable[ref.Registry] = true`. A 404 means the registry IS reachable. Confirmed via smoke test: when two policies share a registry and the first gets a 404, the second is silently dropped from doctor output.

jpower432 (changes requested):
Nice work! Found one bug that was partially pre-existing.

internal/registry/client.go
sonupreetam: It was intentional as part of the error-flattening effort (sub-issue #3), but it over-corrected, non-404 errors lost their module path context. Fixed in 9a87451: `ErrVersionNotFound` passes through unwrapped so `errors.Is` works upstream, while network failures retain the `"failed to fetch version for %s"` wrapping.

internal/doctor/doctor.go
sonupreetam: Good catch,  confirmed the silent-drop behavior. Fixed in 9a87451 by adding `&& !errors.Is(err, registry.ErrVersionNotFound)` to the unreachable guard. Also added `TestCheckPolicyVersions_LatestMissing_DoesNotPoisonSameRegistry` which reproduces the exact scenario (two policies on the same registry, first gets 404, second must still be checked). Three additional edge-case tests added: pinned+both 404, mixed registries (unreachable vs 404), and pinned network failure.

hbraswelrh (approved):
LGTM, but needs a rebase before merge. It can be completed in the UI "Update with rebase"

jpower432 (approved):
LGTM

PR #515, opened 2026-05-18
fix(scan): check evaluator artifacts exist before skipping generation

### LINKED ISSUE #516: fix(scan): stale generation cache skips Generate when artifacts are missing
## Problem

`complyctl scan` silently skips the Generate step when the generation state file (`.complytime/generation/<policy>.json`) exists with a matching digest, even if the actual provider artifacts (e.g., `.complytime/ampel/policy/`) have been deleted. This causes the downstream provider scan to fail with a confusing error.

## Root Cause

`needsRegeneration()` in `cmd/complyctl/cli/scan.go` only compares policy digests via `IsFresh()` but never verifies that the generated artifacts actually exist on disk. If the generation state survives but artifacts are cleaned, complyctl prints "Reusing generated artifacts" and skips the Generate RPC.

## Expected Behavior

Before declaring artifacts fresh, verify that each evaluator's workspace directory (`.complytime/<evaluator>/`) exists and is non-empty. If artifacts are missing, force regeneration.

## Fix

PR #515

Companion issue in complytime-providers: see linked provider-side fix for better error messages when the policy file is missing.

sonupreetam:
## Summary

- Harden `needsRegeneration()` in `cmd/complyctl/cli/scan.go` to verify that evaluator workspace directories (`.complytime/<evaluator>/`) exist and are non-empty before declaring generated artifacts fresh.
- Prevents silent skip of the Generate RPC when the generation state file survives but the actual provider artifacts were deleted.

Fixes #516

## Context

When `complyctl scan` runs, it checks a persisted generation state file to decide whether to skip the Generate step. The freshness check only compared policy digests. If a user (or cleanup) deleted `.complytime/ampel/policy/` but left `.complytime/generation/` intact, complyctl would say "Reusing generated artifacts" and skip Generate, causing the downstream scan to fail with a confusing provider error.

Companion PR in complytime-providers: https://github.com/complytime/complytime-providers/pull/19

## Test plan

- [x] `go build ./cmd/complyctl/cli/` compiles
- [x] `go vet ./cmd/complyctl/cli/` passes
- [x] `go test -race ./cmd/complyctl/cli/` passes
- [x] Pre-push hooks pass (go-unit-tests, go fmt)
- [ ] CI checks pass

### DIFF
--- a/cmd/complyctl/cli/scan.go
+++ b/cmd/complyctl/cli/scan.go
@@ -465,8 +465,27 @@ func needsRegeneration(genState *policy.GenerationState, digest, eid string) boo
 		fmt.Fprintf(os.Stderr, "Policy %s updated since last generate — regenerating\n", eid)
 		return true
 	}
+	if !evaluatorArtifactsExist(genState.EvaluatorIDs) {
+		fmt.Fprintf(os.Stderr, "Generated artifacts missing on disk for %s — regenerating\n", eid)
+		return true
+	}
 	fmt.Fprintf(os.Stderr, "Reusing generated artifacts for %s (policy unchanged)\n", eid)
 	return false
+}
+
+func evaluatorArtifactsExist(evaluatorIDs []string) bool {
+	for _, evalID := range evaluatorIDs {
+		evalDir := filepath.Join(".", complytime.WorkspaceDir, evalID)
+		info, err := os.Stat(evalDir)
+		if err != nil || !info.IsDir() {
+			return false
+		}
+		entries, err := os.ReadDir(evalDir)
+		if err != nil || len(entries) == 0 {
+			return false
+		}
+	}
+	return true
 }
 
 func runGeneration(ctx context.Context, mgr *provider.Manager, groups map[string]policy.EvaluatorGroup, policyTargets []complytime.TargetConfig, globalVars map[string]string, repository, policyDigest string, evaluatorIDs []string) error {
--- a/cmd/complyctl/cli/cli_test.go
+++ b/cmd/complyctl/cli/cli_test.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path/filepath"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


cmd/complyctl/cli/scan.go
```diff
@@ -426,10 +426,29 @@ func needsRegeneration(genState *policy.GenerationState, digest, eid string) boo
...
+		if err != nil || !info.IsDir() {
+			return false
+		}
+		entries, err := os.ReadDir(evalDir)
+		if err != nil || len(entries) == 0 {
+			return false
+		}
+	}
+	return true
+}
```
marcusburghardt, line 489: **[HIGH] Missing tests for new function (TC-001, TC-006)**

This function has 0% test coverage, which causes the CRAP Load CI gate to fail (new functions must stay under CRAP threshold 30; with zero coverage any non-trivial function exceeds it).

TC-006 requires bug fixes to include a regression test that reproduces the original failure. Minimum tests needed (using the existing `chdirTemp()` pattern from `cli_test.go`):

- `TestEvaluatorArtifactsExist_AllPresent` — dirs exist and non-empty → `true`
- `TestEvaluatorArtifactsExist_MissingDir` — dir does not exist → `false`
- `TestEvaluatorArtifactsExist_EmptyDir` — dir exists but empty → `false`
- `TestEvaluatorArtifactsExist_EmptySlice` — no evaluator IDs → `true` (edge case)

Also add `TestNeedsRegeneration_StaleArtifacts` to cover the new code path in `needsRegeneration()` (state file fresh but artifacts missing → returns `true`).

cmd/complyctl/cli/scan.go
```diff
@@ -426,10 +426,29 @@ func needsRegeneration(genState *policy.GenerationState, digest, eid string) boo
...
+		fmt.Fprintf(os.Stderr, "Generated artifacts missing on disk for %s — regenerating\n", eid)
+		return true
+	}
 	fmt.Fprintf(os.Stderr, "Reusing generated artifacts for %s (policy unchanged)\n", eid)
 	return false
 }
 
+func evaluatorArtifactsExist(evaluatorIDs []string) bool {
+	for _, evalID := range evaluatorIDs {
+		evalDir := filepath.Join(".", complytime.WorkspaceDir, evalID)
```
marcusburghardt, line 478: **[MEDIUM] Testability — consider accepting a base directory parameter (AP-002)**

Hardcoding `"."` as the base directory means tests must use `os.Chdir()` to exercise this function. This is consistent with the existing pattern in this file (e.g., line 365), so it is not a regression.

However, accepting a `baseDir string` parameter would eliminate the `chdir` requirement in tests and make the function easier to compose. Consider for a follow-up if you prefer to keep this PR focused on the bug fix.

marcusburghardt (changes requested):
## Summary

Small, focused bug fix that adds a filesystem check to prevent `complyctl scan`
from silently skipping artifact generation when state files survive but actual
provider directories are deleted. The fix logic is correct and well-scoped.

**Blocking issues**: Missing unit tests for both `evaluatorArtifactsExist()` and
the new `needsRegeneration()` code path. This causes the CRAP Load CI gate to
fail (1 regression, 1 new function violation) and violates TC-001 (new
functionality must have tests) and TC-006 (bug fixes must include a regression
test).

marcusburghardt (approved):
LGTM. Thanks for the updates @sonupreetam . It seems we only need to rebase. : )

gvauter (approved):
LGTM

PR #537, opened 2026-06-02
feat: add oci-layout:// scheme for local OCI Layout policy sync

hbraswelrh:
## Summary

Add `oci-layout://` URL scheme support to `complyctl get` so policies can be synced from a local OCI Layout directory instead of requiring a remote OCI registry. This enables demos and private policy evaluation without pushing content to GitHub or any registry.

## Problem

Private policy bundles cannot be committed to GitHub or pushed to a public registry. The only way to use `complyctl get` today is through a remote OCI registry (or the mock registry, which requires embedded testdata at compile time).

## Solution

Introduce `oci-layout:///path/to/bundle` as a new URL scheme in `complytime.yaml`:

```yaml
policies:
  - url: oci-layout:///home/user/bundles/my-private-policy
    id: my-private-policy
```

```bash
complyctl get      # syncs from local OCI Layout, no network
complyctl generate # works unchanged (reads from cache)
complyctl scan     # works unchanged (reads from cache)
```

### Changes

| File | Change |
|------|--------|
| `internal/complytime/config.go` | `OCILayoutScheme` const, `oci-layout://` handling in `ValidateOCIRef` and `ParsePolicyRef` |
| `internal/cache/source.go` | `LocalSource` struct implementing `PolicySource` with `DefinitionVersion` and `CopyPolicy` using `oras.Copy()` local-to-local |
| `cmd/complyctl/cli/get.go` | Scheme dispatch in `syncSinglePolicy`: `oci-layout://` -> `LocalSource`, else -> `RegistrySource` |

### Key Design Decisions

1. **`oci-layout://` scheme** -- Explicit about expected format (valid OCI Layout). `file://` rejected as too ambiguous.
2. **Sentinel dispatch** -- `ref.Registry == "oci-layout://"` routes to `LocalSource`. Minimal change surface.
3. **`oras.Copy()` local-to-local** -- Same transfer mechanism as remote, providing digest verification and incremental sync.
4. **No downstream changes** -- `generate` and `scan` read from `~/.complytime/policies/{id}/` regardless of source.

### Testing

- 10 unit tests for `ValidateOCIRef`/`ParsePolicyRef` with `oci-layout://`
- 8 unit tests for `LocalSource` (happy path, error cases, edge cases)
- 3 E2E tests (local sync, mixed sources, incremental sync)
- All existing tests pass unchanged

### No Breaking Changes

All existing registry-based behavior is unchanged. The `oci-layout://` scheme is purely additive.

## OpenSpec Artifacts

Spec artifacts at `openspec/changes/local-oci-layout-source/` (proposal, design, specs, tasks -- all complete).

### DIFF
--- a/cmd/complyctl/cli/get.go
+++ b/cmd/complyctl/cli/get.go
@@ -109,21 +109,33 @@ func syncSinglePolicy(ctx context.Context, cacheMgr *cache.Cache, state *cache.S
 	ref := complytime.ParsePolicyRef(entry.URL)
 	version := ref.Version
 
-	client := registry.NewClient(ref.Registry, credFunc)
-	source := cache.NewRegistrySource(client)
-	sync := cache.NewSync(cacheMgr, state, source)
+	// Dispatch based on scheme: local OCI Layout vs. remote registry.
+	// For local sources, the cache key is the effective ID (not the path).
+	// For remote sources, the cache key is the repository path.
+	var source cache.PolicySource
+	var syncPolicyID string
+	if ref.Registry == complytime.OCILayoutScheme {
+		source = cache.NewLocalSource(ref.Repository)
+		syncPolicyID = entry.EffectiveID()
+	} else {
+		client := registry.NewClient(ref.Registry, credFunc)
+		source = cache.NewRegistrySource(client)
+		syncPolicyID = ref.Repository
 
-	if version == "" {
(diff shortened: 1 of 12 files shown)


hbraswelrh:
Closing in favor of a devcontainer-only approach that pre-populates the cache directly, avoiding an upstream code change to complyctl.

PR #576, opened 2026-06-10
fix(ci): harden workflows against zizmor security findings

marcusburghardt:
## Summary

Several non-synced workflows in complyctl are missing `persist-credentials: false` on their `actions/checkout` steps, causing zizmor `artipacked` findings that fail the MegaLinter CI check. This is currently blocking dependabot PRs #570 and #571.

This PR:

- Adds `persist-credentials: false` to all `actions/checkout` usages in 6 non-synced workflow files (8 checkout steps total)
- Removes `ci_crapload_main.yml`, which contained both an `artipacked` finding and a `template-injection` vulnerability (`${{ github.ref_name }}` used directly in a `run:` block)

### Files modified

| File | Change |
|---|---|
| `ci_sonarcloud.yml` | Add `persist-credentials: false` |
| `behavioral_assessment.yml` | Add `persist-credentials: false` |
| `e2e_test.yml` | Add `persist-credentials: false` |
| `integration_test.yml` | Add `persist-credentials: false` |
| `release.yml` | Add `persist-credentials: false` |
| `unit_test.yml` | Add `persist-credentials: false` (2 jobs) |
| `ci_crapload_main.yml` | **Deleted** |

### Files NOT touched

Synced workflows (`ci_checks`, `ci_crapload`, `ci_dependencies`, `ci_scheduled`, `ci_security`) are managed by org-infra and already hardened. `ci_cross_repo_integration.yml` already has `persist-credentials: false` on both checkouts. `ci_compliance.yml` has no checkout (calls a reusable workflow only).

## Related Issues

- Unblocks #570 and #571 (dependabot PRs failing on zizmor findings)

## Review Hints

- Each workflow change is a 2-line addition (`with:` + `persist-credentials: false`) except `release.yml` which appends to an existing `with:` block.
- The `ci_crapload_main.yml` deletion is the only structural change. All other files are additive.

PR #582, opened 2026-06-12
fix(proto): resolve protolint line-length violations

yvonnedevlinrh:
## Summary

Wrap long comments in `api/plugin/plugin.proto` to stay within the 80-character protolint limit. Add `protolint:disable` directives for the `go_package` option line (cannot be shortened) and the `CONFIDENCE_LEVEL_NOT_SET` enum value (deliberate naming choice per spec R29 for 1:1 mapping with go-gemara, not an `_UNSPECIFIED` case).

Apply the same fixes to the spec contract proto.

Generated Go code is unchanged - all fixes are proto comments only.

This failure was pre-existing on main.

## Related Issues

- Follow-up to #576 (zizmor security hardening exposed the pre-existing protolint failures in CI)

## Review Hints

- Only two `.proto` files are changed. No generated code, no Go source, no behavioral changes.
- The `protolint:disable:next` directives follow the same pattern already used in `specs/001-gemara-native-workflow/contracts/plugin.proto:10` for the `go_package` line.
- The `CONFIDENCE_LEVEL_NOT_SET` name is intentionally preserved per spec R29 (`specs/001-gemara-native-workflow/research.md:342`) - go-gemara uses `Undetermined` (iota 0) as its zero value, while the proto uses `NOT_SET` (value 0) to distinguish "no confidence reported" from "confidence evaluated but undetermined."

### DIFF
--- a/specs/001-gemara-native-workflow/contracts/plugin.proto
+++ b/specs/001-gemara-native-workflow/contracts/plugin.proto
@@ -33,7 +33,8 @@ message GenerateRequest {
   // Workspace-scoped config (e.g., scan output directory)
   map<string, string> global_variables = 1;
 
-  // Per-requirement test configuration (test variables from policy assessment plan)
+  // Per-requirement test configuration
+  // (test variables from policy assessment plan)
   repeated AssessmentConfiguration configurations = 2;
 }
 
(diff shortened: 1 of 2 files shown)


specs/001-gemara-native-workflow/contracts/plugin.proto
em-redhat: Looks like this is still breaking the 80 character limit

em-redhat: This one is also above 80 characters, causing MegaLinter to fail

em-redhat (approved):
lgtm, all tests passing now

trevor-vaughan (approved):

marcusburghardt (approved):
LGTM

PR #602, opened 2026-06-22
chore: align PolicyRef with OCI reference semantics

hbraswelrh:
## Summary

Addresses [issue #600](https://github.com/complytime/complyctl/issues/600) — follow-up improvements to bring `PolicyRef` into full alignment with OCI Distribution Spec semantics.

#### 1. Deprecate `@version` notation
- Emit a deprecation warning to stderr when `ParsePolicyRef` encounters a non-digest `@` suffix on a full OCI reference (e.g., `@v1.0`), advising the user to switch to `:tag` syntax.
- Warning is deduplicated per URL per process invocation to avoid spam from multiple callers (`EffectiveID`, `validatePolicyRefs`, doctor checks, CLI commands).
- Bare policy IDs (no slash) are exempt since `@` is a complytime convention there, not an OCI digest separator.

#### 2. Split `PolicyRef.Version` into `Tag` and `Digest` fields
- Replace the single `Version` field with separate `Tag string` and `Digest string` fields on `PolicyRef`.
- Add `VersionString()` accessor for backward compatibility (prefers `Tag` over `Digest`).
- Update `ParsePolicyRef` to classify references using oras-go and `go-digest`.
- Update `BuildLookupRef` to accept separate tag and digest parameters.
- Update all callers that read `ref.Version` to use `ref.Tag`, `ref.Digest`, or `ref.VersionString()`.

#### 3. Consistent error reporting in `doctor.go`
- `CheckPolicyActivePeriod` and `CheckComplypacks` now report `ParsePolicyRef` failures as `CheckResult{Status: StatusFail, Blocking: true}`, matching the pattern in `CheckPolicyVersions`.
- `CheckVariables` now emits per-failure `CheckResult{Status: StatusWarn, Blocking: false}` results with specific error messages for four failure cases (policy not found, parse error, version resolution, graph resolution) instead of silently incrementing a counter.

## Related Issues

- Closes complytime/complyctl#600

## Review Hints

- 3 new deprecation warning tests (stderr capture via `os.Pipe` with `defer` protection and `io.ReadAll`)
- 6 new doctor error reporting tests + 2 updated existing tests
- All tests pass with `-race` flag
- `make lint` — 0 issues

### Review Council

Reviewed by 5 Divisor agents. All returned APPROVE after iteration 1 fixes:
- **divisor-adversary**: APPROVE (no security concerns)
- **divisor-architect**: APPROVE (9/10 alignment score)
- **divisor-guard**: APPROVE (aligned with issue intent)
[...]

PR #525, opened 2026-05-21
docs: update policy URLs to per-bundle repo format with policies- prefix

sonupreetam:
## Summary

- Updates `complytime.yaml` to pull policies from `quay.io/complytime/policies-ampel-branch-protection:latest` instead of localhost
- Updates `docs/QUICK_START.md` examples to use new per-bundle OCI repository naming
- Adds cross-reference ADR (`docs/adr/0001-one-quay-repo-per-bundle.md`) pointing to the canonical decision in `complytime-policies`

## Context

Follows from [complytime-policies#36](https://github.com/complytime/complytime-policies/pull/36) (merged) which refactors OCI publishing to use one Quay repository per policy bundle with a `policies-` prefix.

**Before:** `quay.io/complytime/complytime-policies@ampel-branch-protection`
**After:** `quay.io/complytime/policies-ampel-branch-protection:latest`

All three per-bundle repos are now live and publicly accessible on Quay:
- `quay.io/complytime/policies-ampel-branch-protection:latest`
- `quay.io/complytime/policies-cis-fedora-l1-workstation:latest`
- `quay.io/complytime/policies-cis-fedora-l1-server:latest`

## Merge order

| Order | Repo | PR | Purpose | Depends on |
|-------|------|----|---------|------------|
| 1 | org-infra | [#277](https://github.com/complytime/org-infra/pull/277) | Fix reusable compliance workflow + Quay direct fetch | — |
| 2 | complytime-providers | [#22](https://github.com/complytime/complytime-providers/pull/22) | OpenSCAP configuration docs | — |
| **3** | **complyctl** | **#525 (this)** | **Update policy URLs in QUICK_START + complytime.yaml** | #277 merged (for CI) |
| 4 | complytime-policies | [#33](https://github.com/complytime/complytime-policies/pull/33) | Usage docs with provider table + further reading | #22 merged (for doc links) |

## Test plan

- [ ] `complyctl get` resolves the new Quay URLs correctly
- [ ] Quick Start instructions are accurate for new users
- [ ] CI compliance workflow passes after org-infra#277 is merged and re-pinned

Relates to - https://github.com/complytime/complytime-policies/issues/5
Closes https://github.com/complytime/complytime-policies/issues/23

PR #621, opened 2026-06-25
chore: consume complypack from quay.io

marcusburghardt:
## Summary

Update the complypack URL in `.complytime/complytime.yaml` to consume the
released complypack artifact from quay.io instead of the local mock registry.

## Changes

- Replace `http://localhost:8765/complypacks/ampel-bp@v1.0.0` with
  `quay.io/complytime/complypack-ampel-branch-protection:latest`
- Uses `:tag` syntax consistent with the policy URL and the new OCI
  reference standard (the `@version` notation is now deprecated per #602)

## Context

The policy URL was already updated to `quay.io` in #525, but the complypack
entry was still pointing to the mock registry (it was added later as part of
the complypack-pull feature). The complypack artifact is published at
`quay.io/complytime/complypack-ampel-branch-protection` with tags
`latest` and `v0.4.0`.

### DIFF
--- a/.complytime/complytime.yaml
+++ b/.complytime/complytime.yaml
@@ -6,7 +6,7 @@ policies:
     id: ampel-bp
 
 complypacks:
-  - url: http://localhost:8765/complypacks/ampel-bp@v1.0.0
+  - url: quay.io/complytime/complypack-ampel-branch-protection:latest
     id: ampel-bp-pack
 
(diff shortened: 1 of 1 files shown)


hbraswelrh (approved):
LGTM.

yvonnedevlinrh (approved):
LGTM

PR #624, opened 2026-06-25
docs: clarify assessment plan mapping IDs

### LINKED ISSUE #622: Naming: AssessmentConfiguration.RequirementID holds plan ID, not requirement ID
## Summary

`AssessmentConfiguration.RequirementID` in `pkg/provider/client.go` holds the assessment **plan ID**, not the Gemara requirement-id. This is correct behavior for the data flow (Log → Plan → Requirement → Control), but the naming is misleading and caused significant debugging time.

## Details

`ExtractAssessmentConfigs` in `internal/policy/assessment.go` sets:

```go
RequirementID: a.ID,  // plan ID, NOT a.RequirementID (Gemara requirement-id)
```

This is intentional — `a.ID` is the plan ID, and the provider matches on plan IDs. But:

1. The field is called `RequirementID`, suggesting it should hold the Gemara requirement-id
2. `AssessmentConfiguration` has a separate `PlanID` field that goes unused for matching
3. The OPA provider's `MappingEntry` struct also calls the field `RequirementID` and the doc says "maps a Gemara assessment plan RequirementID to a Rego namespace"
4. The `complytime-mapping.json` field is called `requirement_id`

The downstream effect: anyone writing a `complytime-mapping.json` will use Gemara requirement-ids (because the field says "requirement_id"), get zero matches, and see `"ids": null` in scan-config.json. Diagnosing this required adding a debug step to dump scan-config contents and tracing through `ExtractAssessmentConfigs` → `MatchRequirements` to understand which ID is actually used.

## Suggestion

Either:
- Rename `RequirementID` → `PlanID` throughout the chain (provider client, mapping entry, mapping file field), or
- Use the existing `PlanID` field on `AssessmentConfiguration` for matching and keep `RequirementID` for the Gemara requirement-id, or
- At minimum, document that `requirement_id` in the mapping file corresponds to the assessment plan's `id` field, not its `requirement-id` field

sanmaxdev:
## Summary
- Clarifies that OPA complypack `requirement_id` values use Gemara assessment plan `id` values.
- Updates the provider configuration comment to distinguish provider generation IDs from scan output requirement IDs.

## Related Issues
- Closes #622

## Review Hints
- `go test ./internal/policy ./pkg/provider`
- `make build`
- `make lint`

### DIFF
--- a/docs/TESTING_ENVIRONMENT.md
+++ b/docs/TESTING_ENVIRONMENT.md
@@ -245,6 +245,12 @@ Note: The OPA provider requires the `complyctl-provider-opa`
 binary in `~/.complytime/providers/` (installed by the
 post-create script from `complytime-providers`).
 
+For OPA complypacks, `complytime-mapping.json` entries use the
+Gemara assessment plan `id` in `requirement_id`. This is the ID
+sent to providers during generation. It is different from the
+assessment plan's `requirement-id`, which complyctl resolves later
+when writing scan results.
+
 ## Private Bundles
 
 The devcontainer can serve private Gemara policies through
--- a/pkg/provider/client.go
+++ b/pkg/provider/client.go
@@ -24,7 +24,10 @@ type GenerateRequest struct {
 	ComplypackContentPath string
 }
 
-// AssessmentConfiguration binds a requirement ID to its plan and parameters.
+// AssessmentConfiguration carries the assessment plan selected for provider
+// generation. RequirementID is the Gemara assessment plan id used for matching
+// provider content; it is resolved back to the Gemara requirement-id in scan
+// output.
 type AssessmentConfiguration struct {
 	PlanID        string
 	RequirementID string


marcusburghardt (commented):
Hi @sanmaxdev , first of all welcome!! Second, thank you very much for the quick contribution. I think this PR is in a good direction to make things better and it should be fine to merge it. But I also think we have an opportunity to improve the code readability and make it more intuitive. This could be done as part of this PR or in a follow-up PR. So, I would ask if you would like work on this additional improvement? If so we can assign the issue to you and I would be happy to review. :) Thanks

sanmaxdev:
Thanks Marcus, and thanks for the warm welcome.

Yes, I’d be glad to take that improvement. Please assign the issue to me. I’ll keep this PR focused as-is unless you prefer the readability cleanup here, then I can open a follow-up PR for the naming/matching cleanup.

marcusburghardt:
> Thanks Marcus, and thanks for the warm welcome.
> 
> Yes, I’d be glad to take that improvement. Please assign the issue to me. I’ll keep this PR focused as-is unless you prefer the readability cleanup here, then I can open a follow-up PR for the naming/matching cleanup.

That is great. Fine. : )

marcusburghardt (approved):
LGTM

gxmiranda (approved):
LGTM

PR #632, opened 2026-06-26
fix: separate assessment plan and requirement IDs

### LINKED ISSUE #622: Naming: AssessmentConfiguration.RequirementID holds plan ID, not requirement ID

sanmaxdev:
## Summary
- Populates `AssessmentConfiguration.PlanID` from the Gemara assessment plan `id`.
- Keeps `AssessmentConfiguration.RequirementID` as the Gemara `requirement-id`.
- Adds `AssessmentConfiguration.MatchID()` so provider matching prefers `PlanID` and falls back to `RequirementID` when no plan is defined.
- Updates provider test helpers and complypack reference mapping to use the shared match ID behavior.

## Related Issues
- Closes #622
- Follow-up to #624

## Review Hints
- `go test ./internal/policy ./pkg/provider ./cmd/complyctl/cli ./internal/cache ./internal/doctor`
- `make build`
- `make lint`
- `make test-unit`
- `git diff --check`

### DIFF
--- a/cmd/complyctl/cli/scan.go
+++ b/cmd/complyctl/cli/scan.go
@@ -951,15 +951,16 @@ func reverseMap(m map[string]string) map[string]string {
 	return r
 }
 
-// buildReqToComplypackRef produces a pre-resolved requirement-ID →
+// buildReqToComplypackRef produces a pre-resolved match-ID →
 // OCI reference (repository@digest) map by composing two lookups:
 //
 //  1. state.Complypacks: evaluator-ID → repository@digest
-//  2. evaluator groups: requirement-ID → evaluator-ID
+//  2. evaluator groups: match-ID → evaluator-ID
 //
-// Resolving the chain here keeps the Evaluator free of evaluator-ID
-// routing concerns and makes it easier to change selection logic later
-// (e.g., per-requirement complypack selection).
+// The match-ID is cfg.MatchID(), which prefers PlanID and falls back to
+// RequirementID. Resolving the chain here keeps the Evaluator free of
+// evaluator-ID routing concerns and makes it easier to change selection
+// logic later (e.g., per-requirement complypack selection).
 func buildReqToComplypackRef(cacheDir string, groups map[string]policy.EvaluatorGroup) map[string]string {
 	m := make(map[string]string)
 	if len(groups) == 0 {
--- a/cmd/complyctl/cli/scan.go
+++ b/cmd/complyctl/cli/scan.go
@@ -992,7 +992,7 @@ func buildReqToComplypackRef(cacheDir string, groups map[string]policy.Evaluator
 			continue
 		}
 		for _, cfg := range group.Configs {
-			m[cfg.RequirementID] = ref
+			m[cfg.MatchID()] = ref
 		}
 	}
 	return m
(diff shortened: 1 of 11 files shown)


sanmaxdev:
Thanks Jennifer, that makes sense. I updated the follow-up to prefer `PlanID` for provider/content matching and fall back to `RequirementID` when no plan is defined.

Re-ran:
- `go test ./internal/policy ./pkg/provider ./cmd/complyctl/cli ./internal/cache ./internal/doctor`
- `make build`
- `make lint`
- `make test-unit`
- `git diff --check`

cmd/complyctl/cli/scan.go
```diff
@@ -948,7 +948,7 @@ func reverseMap(m map[string]string) map[string]string {
 // OCI reference (repository@digest) map by composing two lookups:
 //
 //  1. state.Complypacks: evaluator-ID → repository@digest
-//  2. evaluator groups:  requirement-ID → evaluator-ID
+//  2. evaluator groups: requirement-ID → evaluator-ID
```
marcusburghardt: **[MEDIUM] Doc comment still says "requirement-ID" but code now uses `cfg.MatchID()`**

The function name and GoDoc describe a requirement-ID → OCI reference mapping, but the code now keys on `MatchID()` which returns plan IDs when present. Worth updating the comment to reflect the match-ID semantics in a follow-up.

marcusburghardt (commented):
Clean, well-scoped fix. All local checks pass (build, lint, unit tests). One minor doc comment to update as a follow-up.

cmd/complyctl/cli/scan.go
sanmaxdev: Good catch. Updated the doc comment on `buildReqToComplypackRef` to say "match-ID" instead of "requirement-ID", and added a note that the match-ID is `cfg.MatchID()`, which prefers `PlanID` and falls back to `RequirementID`.

gvauter (approved):
LGTM

marcusburghardt (approved):
LGTM. Thanks @sanmaxdev

PR #640, opened 2026-06-29
fix: separate assessment log plan IDs

### LINKED ISSUE #635: Naming: AssessmentLog.RequirementID carries plan IDs from providers before resolution
## Summary

`AssessmentLog.RequirementID` in `pkg/provider/client.go` holds plan IDs (via `MatchID()`) when returned by providers during scan. The `resolveAssessmentIDs` function in `cmd/complyctl/cli/scan.go` maps these back to actual requirement IDs before output, so correctness is not affected.

However, the field name is misleading — it says "RequirementID" but carries a plan ID at that point in the data flow. This is the same category of naming confusion that #622 addressed for `AssessmentConfiguration`, just one layer deeper on the response path.

## Related

- Follow-up from #632
- Similar to #622

sanmaxdev:
## Summary
- Adds `PlanID` to `AssessmentLog` for provider-returned match IDs before scan output resolution.
- Keeps `RequirementID` as the final Gemara requirement-id used by output/reporting code.
- Updates scan ID resolution, provider server/client mapping, test providers, and regression tests for the response path.

## Related Issues
- Closes #635
- Follow-up to #632

## Review Hints
- `go test ./internal/policy ./pkg/provider ./cmd/complyctl/cli ./internal/cache ./internal/doctor -count=1`
- `make build`
- `make lint`
- `TMPDIR=/tmp make test-unit`
- `git diff --check`

### DIFF
--- a/cmd/complyctl/cli/cli_test.go
+++ b/cmd/complyctl/cli/cli_test.go
@@ -767,17 +767,18 @@ func TestExtractPlanToReqMap_EmptyAssessments(t *testing.T) {
 
 func TestResolveAssessmentIDs_ReplacesMatchingIDs(t *testing.T) {
 	assessments := []provider.AssessmentLog{
-		{RequirementID: "ap-1"},
+		{PlanID: "ap-1"},
 		{RequirementID: "req-already-correct"},
 	}
 	planToReq := map[string]string{"ap-1": "req-1"}
 	resolveAssessmentIDs(assessments, planToReq)
+	assert.Equal(t, "ap-1", assessments[0].PlanID)
 	assert.Equal(t, "req-1", assessments[0].RequirementID)
 	assert.Equal(t, "req-already-correct", assessments[1].RequirementID)
 }
 
 func TestResolveAssessmentIDs_EmptyMap(t *testing.T) {
-	assessments := []provider.AssessmentLog{{RequirementID: "ap-1"}}
+	assessments := []provider.AssessmentLog{{PlanID: "ap-1"}}
 	resolveAssessmentIDs(assessments, map[string]string{})
 	assert.Equal(t, "ap-1", assessments[0].RequirementID)
 }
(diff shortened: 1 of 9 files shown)


trevor-vaughan:
Hi @sanmaxdev , thanks for the PR!

I was digging through and noticed the one change in `manager_test.go` but I think that a few other spots may have been missed.

Lines 95 and 237 still seem to populate `RequirementID` but I believe that these should be switched over to `PlanID` with your fixes.

Likewise, I think that the assertions on 260 and 351 should also be checking `PlanID` now, correct?

On line 318 of `manager_test.go`, I noticed that we're still using `RequirementID` instead of `PlanID`, was this a deliberate choice? If not, should we test both the new primary as `PlanID` and the fallback?

Finally, I noticed that `errorAssessments()` in `pkg/provider/manager.go` appears to log `RequirementID:` as a field. I did a bit of backtracing but think that this might not be correct given that the field can be populated by a few different avenues. We may need to keep it for backwards compatibility but I'd like to get your opinion if you're up for it.

yvonnedevlinrh (dismissed):
LGTM

sanmaxdev:
Pushed a follow-up for the scan match ID tests. Provider-returned assessments now assert `PlanID`, and the broadcast test covers both the `PlanID` path and the `RequirementID` fallback path.

Verification:
- `go test -mod=vendor ./pkg/provider -run 'TestManager_RouteScanResult_PartialResults|TestManager_RouteScanResult_Broadcast_RPCFailure|TestManager_RouteScan_DropsProviderErrors|TestManager_RouteScan_SpecificProvider' -count=1`
- `go test -mod=vendor ./pkg/provider -count=1`
- `go test -mod=vendor ./pkg/provider -race -count=1`
- `go vet ./pkg/provider`
- `git diff --check`

marcusburghardt:
LGTM. The second commit addresses trevor-vaughan's feedback comprehensively. Just needs a rebase to resolve the merge conflict.

marcusburghardt (dismissed):
LGTM

sanmaxdev:
Rebased on current main and fixed the goimports cleanup from the rebase.

Verification:
- `make test-unit`
- `make lint`
- `git diff --check`

trevor-vaughan (approved):
LGTM

gvauter (approved):
Clean, well-scoped fix that adds a `PlanID` field to `AssessmentLog` to separate the provider-returned match ID from the final resolved Gemara requirement ID.

The resolution logic in `resolveAssessmentIDs` correctly handles all three cases: PlanID-only (new path), RequirementID-only (backward compat), and both set (PlanID takes precedence). Tests cover each case. All CI checks pass.

PR #641, opened 2026-06-29
feat: add policy verification pipeline and list integrity display

### LINKED ISSUE #607: `get` pulls policy bundles without verifying signatures (THR02 mitigation gap)
## Summary

`complyctl get` fetches policy artifacts over OCI with transfer-integrity only — no
signature verification and no warning — even though the project's own threat catalog names
mutable-tag/registry tampering (THR02) as a top supply-chain threat and the published
policy bundles are already signed. Because policies determine pass/fail verdicts, an
unverified policy layer is a direct integrity compromise of the assessment itself.

## Affected

- `internal/cache/source.go:41`
- `cmd/complyctl/cli/get.go` (policy pull path; complypack warning at line 235)
- Threat catalog: `governance/threats/complytime-threats.yaml:57-66` (THR02)

## Detail

THR02, in the catalog's own words:

> "Policy artifacts fetched by mutable tags (e.g., "latest") without content digest pinning
> allow a compromised or man-in-the-middle registry to serve malicious catalog, guidance, or
> policy layers. The cache sync stores whatever the registry serves without verifying content
> against a known-good digest."

The policy pull does not verify signatures:

```go
// internal/cache/source.go:41
oras.Copy(ctx, repo, tag, dst, tag, oras.CopyOptions{})
```

This provides digest/transfer integrity only — no signature check, no warning. The
complypack path at least warns at runtime
(`WARNING: complypack … has not been cryptographically verified`,
`cmd/complyctl/cli/get.go:235`), but the policy path is silent.

Meanwhile the documented/dogfooded configs steer users onto mutable tags: the dogfooded
config (`complytime-policies/.complytime/complytime.yaml`) pins all three policies with
`@latest`, and the deprecation nudge toward `:tag` (tracked separately) is still a mutable
tag, still inside THR02. So the project's own #1 supply-chain threat is unmitigated by
default on the exact path the docs push users onto.

## Why signature verification, not digest pinning

Digest pinning freezes content (a staleness footgun) and does not stop a malicious registry
that serves a self-consistent bad bundle. THR02 is a tampering/MITM threat; the mitigation
that addresses it without freezing content is verifying the **signature** on the pulled
bundle. `complytime-policies` already signs bundles to Quay (sigstore) — the same model the
release binaries use — so `get` can verify the signing identity/cert on every pull
regardless of whether the tag is mutable. That delivers integrity **and** freshness under
mutable tags.

## Suggested fix

- Wire signature verification into `get` for both policies and complypacks (this depends on
  finishing the sigstore-go integration tracked separately against `complypack`).
- Until verification is enforced, warn on the policy pull path the way the complypack path
  already does.
- Optionally surface the resolved digest in `complyctl list` so an org that wants
  reproducibility can pin and re-pin deliberately.

This is the `complyctl` (wire-in + policy-path warning) half of the THR02 work; the
underlying `verify()` implementation is tracked in a separate `complypack` issue.

---

**Related:** complytime/complypack#63, complytime/website#62, complytime/complyctl#612

Issue created with the help of Claude Opus 4.6

hbraswelrh:
## Summary

Addresses #607 — `get` pulls policy bundles without verifying signatures (THR02 mitigation gap).

This PR adds user-facing visibility into artifact integrity: unverified warnings on fresh pulls, a DIGEST column in `complyctl list`, and THR02 mitigation documentation.

The verification infrastructure (`VerifyFunc`, `SyncOption`, `Verified` state field, VERIFIED column) was de-scoped per review feedback from @marcusburghardt and deferred to #643, where it will be designed against the actual sigstore-go API to avoid premature schema commitments before the stable release.

### Policy Sync Return Type (`internal/cache/sync.go`)
- `SyncPolicy` returns `(bool, error)` instead of `error` — `true` indicates a fresh fetch occurred, enabling the CLI to gate warnings on new artifacts vs incremental skips

### Complypack Sync (`internal/cache/complypack_sync.go`)
- `SyncComplypack` continues returning `(bool, error)` (unchanged from main) — the `true` return gates the complypack warning

### CLI Warning (`cmd/complyctl/cli/get.go`)
- `complyctl get` emits `WARNING: policy <id> has not been cryptographically verified` to stderr when a policy is freshly fetched
- `complyctl get` emits `WARNING: complypack <id> has not been cryptographically verified` to stderr when a complypack is freshly fetched
- Warnings are only emitted on fresh fetches, not incremental skips

### List Command (`cmd/complyctl/cli/list.go`)
- `complyctl list` displays a DIGEST column showing the abbreviated OCI manifest digest (e.g., `sha256:9f86d081884c`)
- Uncached policies show `-` for the digest
- Cache state is loaded via `LoadState` to populate the digest field

### Threat Documentation (`governance/threats/complytime-threats.yaml`)
- THR02 gains 2 mitigation entries:
  - MIT01: unverified warning on fresh pulls (implemented)
  - MIT03: DIGEST column visibility (implemented)

## Testing

- Existing sync tests updated for `(bool, error)` return type
- New list tests: `TestListOptions_Run_ShowsDigestColumn`, `TestListOptions_Run_UncachedPolicyShowsDash`, `TestAbbreviateDigest`
- All changed packages pass: `internal/cache`, `cmd/complyctl/cli`, `internal/policy`
- `make lint`: 0 issues

## Deferred to #643

The following items were removed from this PR and deferred to #643 (sigstore-go verification) per review feedback:

- `verify.go` (`VerifyFunc`, `SyncOption`, `WithVerifier`, `runVerify`, `UnverifiedWarning`)
- `Verified bool` on `PolicyState` and the VERIFIED column
- THR02 mitigations MIT02 and MIT04

The `opsx/sigstore-verification` branch retains the original verification infrastructure as a reference point. Issue #643 has been updated with the deferred scope.

## Upstream Reference Update

`complypack#63` is closed, superseded by `complypack#83` (removing sign/verify stubs from complypack) and #643 (sigstore-go verification in complyctl).

## Closes

Closes #607


cmd/complyctl/cli/get.go
```diff
@@ -159,21 +159,33 @@ func syncSinglePolicy(ctx context.Context, cacheMgr *cache.Cache, state *cache.S
 
 	client := registry.NewClient(ref.Registry, credFunc)
 	source := cache.NewRegistrySource(client)
-	sync := cache.NewSync(cacheMgr, state, source)
+	sync := cache.NewSync(cacheMgr, state, source, cache.NoOpVerifier())
```
trevor-vaughan: Even though we're using the `NoOpVerifier`, we shouldn't expose it directly via the interface. This should call "regular" verification and `NoOp` just happens to be the current implementation.

I probably wouldn't even call out `NoOp` since it'll be logged as such and we'll replace it with the real implementation soon.

cmd/complyctl/cli/get.go
```diff
@@ -159,21 +159,33 @@ func syncSinglePolicy(ctx context.Context, cacheMgr *cache.Cache, state *cache.S
...
+	fetched, err := sync.SyncPolicy(ctx, ref.Repository, version)
+	if err != nil {
 		fmt.Fprintln(os.Stderr, "failed")
 		logger.Error("Policy sync failed", "policy", ref.Repository, "error", err)
 		return err
 	}
 	fmt.Fprintln(os.Stderr, "done")
-	logger.Info("Policy synced", "policy", entry.EffectiveID())
+
+	if fetched {
```
trevor-vaughan, line 179: This logic should probably go into the verifier/cache directly instead of in the caller.

cmd/complyctl/cli/get.go
trevor-vaughan: Hmm....we seem to have some duplication with this. These should definitely be moved into the `cache` since that's where the abstraction lives.

internal/cache/sync.go
```diff
@@ -79,23 +94,39 @@ func (s *Sync) SyncPolicy(ctx context.Context, policyID, version string) error {
...
 
-	return nil
+	return true, nil
+}
+
+// verifyAfterCopy runs the verifier against the copied descriptor.
+// Returns true when the artifact is verified, false otherwise.
+// Verifier errors are treated as unverified (graceful degradation)
+// and logged as warnings for operational visibility.
+func verifyAfterCopy(ctx context.Context, v Verifier, desc ocispec.Descriptor) bool {
```
trevor-vaughan: Given that the payload is decoupled from the signature (if I'm remembering correctly), wouldn't we want to verify before pulling the object? We would then simply refuse to download unverified material based on the intent of the user.

This does mean that we'll need to allow the user to bypass this restriction in case they need to pull an unverified object for whatever reason (broken publish, testing, etc...).

internal/cache/sync.go
```diff
@@ -79,23 +94,39 @@ func (s *Sync) SyncPolicy(ctx context.Context, policyID, version string) error {
...
+	result, err := v.Verify(ctx, desc)
+	if err != nil {
+		slog.Warn("Policy verification failed, treating as unverified",
```
trevor-vaughan: This message should probably move to `verify.go` to reduce duplication/sprawl (same as in `get.go`)

After reduction, this method can probably be eliminated since the verifier will handle everything natively.

trevor-vaughan (changes requested):

gvauter (dismissed):
LGTM

marcusburghardt (changes requested):
Nice work -- the verification pipeline design is solid and the test coverage is thorough. The WARNING on policy pulls and the DIGEST column in `list` are exactly what we need to close the THR02 visibility gap.

Since we're close to a stable release, I'd like to suggest scoping this PR to just those two pieces for now. The reason is risk mitigation: the verification schema (`Verified bool` in `state.json`, VERIFIED column, `VerifyFunc`/`SyncOption` plumbing) introduces external commitments that we'd need to maintain post-stable, and there are still open design questions in #643 that could affect them -- pre-copy vs post-copy timing, what verification metadata to persist (sigstore-go returns rich results beyond a boolean: signer identity, timestamps, Rekor entries), and how `--skip-verify` interacts with state. If any of those decisions change the shape of the schema or CLI surface, we'd be looking at a migration instead of a clean introduction.

By shipping the visibility improvements now and deferring the verification infrastructure to #643, we can design the state schema, plumbing, and VERIFIED column against the actual sigstore-go API in one go -- no migration risk, no post-stable breaking changes.

One housekeeping note: the upstream reference has changed. complypack#63 is closed, superseded by complypack#83 (removing sign/verify stubs from complypack) and #643 (sigstore-go verification in complyctl).

**Suggested scope for this PR:**
- WARNING on fresh policy pulls (parity with the complypack path)
- DIGEST column in `complyctl list`
- `SyncPolicy` returning `(bool, error)` to gate warnings on fresh fetches
- THR02 mitigations MIT01 (warning) and MIT03 (digest visibility)

**Deferred to #643:**
- `verify.go` (`VerifyFunc`, `SyncOption`, `WithVerifier`, `runVerify`)
- `Verified bool` on `PolicyState` and the VERIFIED column
- THR02 mitigations MIT02 and MIT04

Happy to discuss if you see this differently -- the underlying work is great and I want to make sure it lands in the best shape for the stable release.

hbraswelrh:
@marcusburghardt @trevor-vaughan this will require a re-review since changes were requested on your end.

marcusburghardt (dismissed):
LGTM. Thanks for the quick updates @hbraswelrh . 👍

cmd/complyctl/cli/cli_test.go
```diff
@@ -520,6 +520,82 @@ func TestListOptions_Run_ValidWorkspace(t *testing.T) {
...
+}
+
+func TestAbbreviateDigest(t *testing.T) {
+	tests := []struct {
+		name   string
+		input  string
+		expect string
+	}{
+		{"full sha256", "sha256:9f86d081884c7d659a2feaa0c55ad015a3bf4f1b", "sha256:9f86d081884c"},
+		{"short hex", "sha256:abc", "sha256:abc"},
```
trevor-vaughan: This isn't valid, is it a stub/placehoder? We should have a 'skip' on tests that are placeholders so that we know it's not a mistake.

cmd/complyctl/cli/cli_test.go
```diff
@@ -520,6 +520,82 @@ func TestListOptions_Run_ValidWorkspace(t *testing.T) {
...
+		{"empty", "", "-"},
+		{"no colon", "invaliddigest", "invaliddigest"},
+		{"sha512", "sha512:abcdef123456789012", "sha512:abcdef123456"},
```
trevor-vaughan: This isn't valid, is it a stub/placehoder? We should have a 'skip' on tests that are placeholders so that we know it's not a mistake.

cmd/complyctl/cli/get.go
```diff
@@ -167,13 +167,23 @@ func syncSinglePolicy(ctx context.Context, cacheMgr *cache.Cache, state *cache.S
...
+		ps, _ := state.GetPolicyState(ref.Repository)
+		logger.Info("Policy synced", "policy", entry.EffectiveID(), "digest", ps.Digest)
+		fmt.Fprintf(os.Stderr, "WARNING: policy %s has not been cryptographically verified\n", entry.EffectiveID())
```
trevor-vaughan: We're printing a scary-looking warning here and that makes the user think that they should do something about it and/or something is wrong.

Given that this is a limitation in our software, we need to let the user know how to work around this in the message itself. We can either tell them that this will be an upcoming capability (and to ensure that their sources are properly signed) or we can tell them how to use oras + other tooling if they really want to validate the signature.

As it is, I just think "oh no, something's wrong, I should stop using this until it's fixed".

As a last alternative, we could just not emit any message and just rely on the documentation (this is technically better than that though).

trevor-vaughan (changes requested):

cmd/complyctl/cli/cli_test.go
hbraswelrh: Replaced with sha256:9f86d081884c (exact 12 hex chars) which is a valid truncated digest that exercises the same boundary condition (input not longer than 12 chars after the colon).

cmd/complyctl/cli/cli_test.go
hbraswelrh: Replaced with the actual SHA-512 hash of empty input (sha512:cf83e1357eef...), a real digest that exercises the sha512 algorithm prefix path.

cmd/complyctl/cli/get.go
hbraswelrh: Changed from WARNING: policy X has not been cryptographically verified to NOTE: policy X was fetched without signature verification. Signature verification support is planned (see https://github.com/complytime/complyctl/issues/643). Until then, ensure your policy sources are trusted. Applied the same change to the complypack path. Updated threat docs, CHANGELOG, and AGENTS.md for consistency.

trevor-vaughan (approved):

marcusburghardt (approved):
LGTM

PR #652, opened 2026-06-30
chore: add preflight validation gate to release workflow

### LINKED ISSUE #560: chore: add release preflight validation and concurrency protection
## Summary

A comparison of our release workflow with the preflight pattern used in
`unbound-force/gaze` (PR unbound-force/gaze#100) and
`unbound-force/unbound-force` identified several improvement opportunities
to make the release process more robust.

## Current State

The release workflow is a single `workflow_dispatch` job that runs GoReleaser
immediately on dispatch. It relies on branch protection to ensure code quality
before code reaches `main`. This works well but leaves room for additional
safety nets.

### 1. No CI verification
The workflow trusts branch protection to ensure CI passed on the commit being
released. Adding an explicit query to the GitHub Checks API for required check
run conclusions would provide defense in depth — catching edge cases where
branch protection is temporarily adjusted or bypassed by an admin.

### 2. No concurrency protection
No `concurrency:` group is defined. If two `workflow_dispatch` runs are
triggered in quick succession, they could race. A concurrency group would
serialize releases safely.

### 3. No semver ordering check
Nothing prevents releasing `v1.1.0` after `v1.2.0` has already been released,
which could confuse downstream consumers.

### 4. No unreleased commits guard
Nothing prevents triggering a release that produces the same artifacts as the
previous tag (no new commits). A simple `git rev-list --count` check would
catch this early.

### 5. No tag creation by workflow
The workflow uses `GORELEASER_CURRENT_TAG` to name the release but does not
create a git tag. The tag may or may not exist in git history depending on
whether the maintainer also pushed one separately. Having the workflow create
an annotated tag (idempotently) after validation would ensure consistency.

## Suggested Improvements

- [ ] Concurrency group (`release-${{ github.ref }}`, `cancel-in-progress: false`)
- [ ] CI verification via GitHub Checks API (query required check run conclusions on HEAD)
- [ ] Semver ordering check against latest existing tag
- [ ] Unreleased commits guard (`git rev-list --count`)
- [ ] Annotated tag creation by the workflow (idempotent)

## Reference

- `unbound-force/gaze` PR #100: preflight gate implementation
- `unbound-force/unbound-force` release workflow: original pattern

marcusburghardt:
## Summary

Add a `preflight` job to the release workflow that validates
preconditions before building and publishing release artifacts.
Replicates the pattern from `unbound-force/gaze`
([PR #100](https://github.com/unbound-force/gaze/pull/100)).

### Preflight validations (8 steps)

- **Tag format**: `vMAJOR.MINOR.PATCH[-PRERELEASE]` (relaxed for
  v1.0.0 transition; TODO to tighten after)
- **Tag uniqueness**: rejects duplicate tags
- **Semver ordering**: ensures new tag is greater than latest
- **CI verification**: queries GitHub Checks API for `unit-test`,
  `e2e-test`, `integration-test` conclusions on HEAD
- **Unreleased commits guard**: rejects empty releases
- **Annotated tag creation**: idempotent, skips if tag exists

### Other improvements

- **Concurrency group** (`release-${{ github.ref }}`,
  `cancel-in-progress: false`) prevents parallel releases
- **Release job** now depends on preflight (`needs: preflight`)
  and checks out the validated tag ref
- **Workflow header** comment documents the pipeline and links
  to this issue

### CHANGELOG

Entry added under `### Changed` in the `Unreleased` section.

Closes #560

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,3 +1,14 @@
+# Release
+# =======
+# Release pipeline triggered via workflow_dispatch.
+# Validates preflight conditions (semver format, tag
+# uniqueness, CI status, unreleased commits), creates
+# an annotated tag, and builds release artifacts via
+# GoReleaser with cosign signatures and SBOM.
+#
+# Pattern replicated from unbound-force/gaze.
+# See: https://github.com/complytime/complyctl/issues/560
+
 name: Release
 
 on:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -8,25 +19,146 @@ on:
...
+                  LATEST=$(git tag -l 'v[0-9]*' --sort=-v:refname | head -1)
+                  if [ -z "$LATEST" ]; then
+                      echo "No existing tags found. First release."
+                      exit 0
+                  fi
+                  echo "Latest existing tag: $LATEST"
+                  # Compare using sort -V: if TAG sorts after
+                  # LATEST, it is greater.
+                  HIGHER=$(printf '%s\n%s' "$LATEST" "$RELEASE_TAG" | sort -V | tail -1)
+                  if [ "$HIGHER" = "$LATEST" ]; then
```
hbraswelrh, line 78: **[MEDIUM — non-blocking]** `sort -V` handles standard semver well on Ubuntu's GNU coreutils, but its pre-release comparison behavior diverges from the semver spec (e.g., ordering `v1.0.0-alpha.1` vs `v1.0.0-beta.0` may not follow semver §11 precedence rules). Fine for now given the TODO to tighten after v1.0.0 — just flagging for when that hardening pass happens. A small bash semver comparator or a dedicated tool would be more reliable for pre-release ordering long-term.

hbraswelrh (approved):
Looks good. One non-blocking observation on the semver comparison approach for future hardening.

hbraswelrh:
@marcusburghardt ready to approve once the branch is rebased 👍

hbraswelrh (approved):
LGTM.

sonupreetam (approved):
LGTM

PR #661, opened 2026-07-01
fix: complypack cache reliability — version eviction and generation state accuracy

### LINKED ISSUE #645: bug: LookupByEvaluatorID selects arbitrary complypack version when multiple exist
## Problem

`ComplypackCache.LookupByEvaluatorID()` iterates `~/.complytime/complypacks/{evaluator-id}/` via `os.ReadDir` and returns the first version directory found. Directory iteration order is filesystem-dependent and non-deterministic in Go. If multiple version directories coexist (e.g. `0.1.0/` and `0.2.0/`), which content gets passed to the provider is undefined.

This can cause providers to receive outdated complypack content with no error or warning.

## Steps to reproduce

1. Pull complypack v1.0.0 → cached at `~/.complytime/complypacks/opa/1.0.0/`
2. Update complypack in registry to v2.0.0 (different version field in config)
3. `complyctl get` → stores at `~/.complytime/complypacks/opa/2.0.0/` (old `1.0.0/` remains)
4. `complyctl scan` → `LookupByEvaluatorID("opa")` may return `1.0.0` or `2.0.0`

## Expected behavior

Either: (a) only the latest version should exist (evict old on store), or (b) `LookupByEvaluatorID` should deterministically select the most recent version (e.g. by `state.json` metadata or directory mtime).

## Related

- #551 — complypack cache invalidation gap
- `internal/cache/complypack.go` line 190 — `LookupByEvaluatorID`
- Enabling condition: old versions are never cleaned up (see sibling enhancement)

## Acceptance Criteria

- [ ] `ComplypackCache.Store()` removes all sibling version directories for the same evaluator-id before writing the new version
- [ ] Only one version directory exists per evaluator-id after any `Store()` call
- [ ] Other evaluator-id directories are not affected
- [ ] Unit tests covering eviction of single/multiple old versions, idempotent re-store, and cross-evaluator isolation

### LINKED ISSUE #646: enhancement: evict old complypack versions from cache on re-sync
## Summary

When a complypack's `version` field changes between releases (e.g. `0.1.0` → `0.2.0`), `ComplypackCache.Store()` only does `RemoveAll` on the matching version path. The old version directory remains on disk indefinitely. There is no cache eviction, no "keep only latest" logic, and no `complyctl cache clean` command.

## Impact

- Unbounded disk usage in `~/.complytime/complypacks/`
- Enables the non-deterministic version selection bug in `LookupByEvaluatorID` (#645)
- Stale content can be silently used by providers

## Proposed fix

When `ComplypackCache.Store()` writes a new version for an evaluator-id, remove all sibling version directories under `{cacheDir}/complypacks/{evaluator-id}/` that are not the version being written. This ensures only one version exists per evaluator-id at any time.

Alternative: add explicit eviction in `ComplypackSync.SyncComplypack()` after successful store, removing prior version directories for the same evaluator-id.

## Related

- #645 — `LookupByEvaluatorID` picks arbitrary version (sibling bug)
- #551 — complypack cache invalidation gap
- `internal/cache/complypack.go` — `Store()` method

## Acceptance Criteria

- [ ] `Store()` calls `os.RemoveAll` on all sibling version directories (not hidden, not the target version) before atomic rename
- [ ] No error returned if old version directories don't exist
- [ ] Unbounded disk growth from version changes is eliminated
- [ ] Unit tests for eviction behavior

### LINKED ISSUE #648: bug: complyctl generate records generation state from potentially stale state.json
## Problem

`complyctl generate` always invokes the provider Generate RPC (no freshness check), then calls `saveGenerationAndPrint()` which reads `state.json` for the current policy/complypack digests and saves them to the generation state file. If `state.json` has drifted from the actual cache contents (e.g. cache directory manually deleted but state not cleared), the recorded digests won't match what was actually used during generation.

On the next `complyctl scan`, `IsFresh()` compares the (stale) recorded digests against current `state.json` digests. If they match (both stale), scan skips regeneration even though workspace artifacts may be based on different content.

## Steps to reproduce

1. `complyctl get` → state.json records digest A for complypack
2. Manually delete `~/.complytime/complypacks/{evaluator}/` (simulating corruption)
3. `complyctl generate --policy-id X` → provider runs with empty complypack path ("")
4. `saveGenerationAndPrint` reads state.json → still shows digest A → saves to generation state
5. Re-run `complyctl get` → re-fetches same complypack (digest A) → state unchanged
6. `complyctl scan` → digests match, artifacts exist → "Reusing generated artifacts" (but they were generated without complypack content)

## Expected behavior

`saveGenerationAndPrint` should record the digest of what was *actually used* during generation, not what `state.json` currently claims. When complypack content path was empty during generate, the recorded complypack digest should reflect that (empty or absent).

## Related

- `cmd/complyctl/cli/generate.go` — `saveGenerationAndPrint()` (line 165)
- #583 — generation state complypack digest tracking

## Acceptance Criteria

- [ ] `saveGenerationAndPrint` records empty/absent complypack digest when content path was unavailable during generation
- [ ] Next `complyctl scan` triggers regeneration when complypack becomes available (digest mismatch)
- [ ] No change to behavior when complypack content was available (digest recorded normally)
- [ ] Unit tests for mixed availability scenarios

sonupreetam:
## Summary

Fixes three related complypack cache reliability issues:

- **#645**: `LookupByEvaluatorID` selects arbitrary complypack version when multiple exist
- **#646**: Old complypack versions not evicted from cache on re-sync
- **#648**: `complyctl generate` records generation state from potentially stale `state.json`

### Complypack version eviction (#645, #646)

`ComplypackCache.Store()` now evicts all sibling version directories for the same evaluator-id before writing the new version. This ensures at most one version exists per evaluator-id, making `LookupByEvaluatorID()` deterministic.

- Added `evictOldVersions()` helper in `internal/cache/complypack.go`
- Eviction errors are non-fatal (warning logged, store continues)
- Hidden/temp directories (dot-prefixed) are skipped during eviction

### Generation state accuracy (#648)

`saveGenerationAndPrint()` and `runGeneration()` now filter complypack digests to only include evaluator-ids whose content was actually resolved during generation. When complypack content is unavailable (cache deleted, not yet synced), its digest is recorded as empty — forcing regeneration on the next scan when the complypack becomes available.

- `generateForAllTargets()` returns available evaluator-ids
- Added `filterComplypackDigests()` helper
- Both generate and scan code paths apply the same filtering

## Files Changed

| File | Change |
|------|--------|
| `internal/cache/complypack.go` | `evictOldVersions()`, import `charmbracelet/log` |
| `internal/cache/complypack_test.go` | 5 eviction tests, replaced `TestStore_MultipleVersions` |
| `cmd/complyctl/cli/generate.go` | `saveGenerationAndPrint` gains `availableEvaluators` param |
| `cmd/complyctl/cli/scan.go` | `generateForAllTargets` returns available evaluators, `filterComplypackDigests()` |
| `cmd/complyctl/cli/cli_test.go` | 8 new tests (4 filter + 4 integration) |

## Testing

- `go test -race ./internal/cache/` — all 49 tests pass
- `go test -race ./cmd/complyctl/cli/` — all tests pass
- `go vet ./...` — clean
- `make test-integration` — 26/26 pass
- E2E verification with mock registry: version eviction confirmed, generation state accuracy confirmed

## OpenSpec Artifacts

- `openspec/changes/complypack-version-eviction/` — proposal, design, spec, tasks (11/11 complete)
- `openspec/changes/generate-state-from-actual-usage/` — proposal, design, spec, tasks (12/12 complete)

Fixes #645
Fixes #646
Fixes #648

### DIFF
--- a/cmd/complyctl/cli/cli_test.go
+++ b/cmd/complyctl/cli/cli_test.go
@@ -448,8 +448,9 @@ func TestGenerateForAllTargets_EmptyGroups(t *testing.T) {
 	cacheDir := t.TempDir()
 	groups := map[string]policy.EvaluatorGroup{}
 	targets := []complytime.TargetConfig{{ID: "local"}}
-	err := generateForAllTargets(context.Background(), cacheDir, nil, groups, targets, nil)
+	available, err := generateForAllTargets(context.Background(), cacheDir, nil, groups, targets, nil)
 	assert.NoError(t, err)
+	assert.Empty(t, available)
 }
 
 // --- getOptions tests ---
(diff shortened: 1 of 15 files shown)


cmd/complyctl/cli/cli_test.go
```diff
@@ -1059,6 +1060,188 @@ func TestComplypackDigestsByEvaluator_ExtractsDigests(t *testing.T) {
...
 
+// --- filterComplypackDigests tests ---
+
+func TestFilterComplypackDigests_AllAvailable(t *testing.T) {
+	allDigests := map[string]string{
+		"io.complytime.opa":   "sha256:aaa",
+		"io.complytime.ampel": "sha256:bbb",
+	}
+	available := []string{"io.complytime.opa", "io.complytime.ampel"}
+	result := filterComplypackDigests(allDigests, available)
```
trevor-vaughan, line 1072: This has actually been bugging me.

Ideally, `filterComplypackDigests` would reject/filter our any items that are not actually correct.

`sha256:aaa` should not be returned or recognized since it's not a valid `sha256` checksum. If (somehow) something gets here in this state, it should be logged as an internal error and not presented to the user.

In theory, the keys should also be validated but I'm not 100% sure of the valid set.

openspec/changes/complypack-version-eviction/specs/version-eviction/spec.md
```diff
@@ -0,0 +1,35 @@
+## ADDED Requirements
+
+### FR-001: Evict prior versions on complypack store
+
+**Given** a complypack with evaluator-id `E` version `2.0.0` is being stored
+**And** a prior version directory `~/.complytime/complypacks/E/1.0.0/` exists
+**When** `ComplypackCache.Store()` completes successfully
+**Then** the directory `~/.complytime/complypacks/E/1.0.0/` MUST NOT exist
```
trevor-vaughan, line 8: I don't think this is a good default. It's pretty common to want to compare new and old versions of things and auto-eviction is going to be confusing.

Do I have a way to select an older version? What happens if I download a newer version then want to use an older version and switch back during testing or because of a published bug?

A possible solution to the referenced issue would be to human sort based on the semantic version of the downloaded packs and use the highest number as the default.

trevor-vaughan (changes requested):

yvonnedevlinrh (dismissed):
LGTM
Only medium finding which is just a suggestion 
evictOldVersions is unexported but receives an evaluatorDir parameter that could be any path — consider adding a brief defensive comment or assertion that the directory is within the expected cache root. This is a style suggestion for future maintainers; the current callers always pass validated paths. (CS-003)

cmd/complyctl/cli/cli_test.go
sonupreetam: The `sha256:aaa` values here are test fixture placeholders, the same pattern used throughout the existing test suite (e.g., `sha256:abc123` in `sync_test.go`, `sha256:abc` in `needsRegeneration` tests, `sha256:def456` in `buildReqToComplypackRef` tests).

`filterComplypackDigests` is responsible for filtering by evaluator availability, not digest format validation. Digest values arrive from OCI registry responses via `state.json` and are validated at ingestion time during `complyctl get`. Adding format validation here would duplicate that responsibility and is out of scope for this change.

If digest format validation is desired as a defense-in-depth measure, it would belong in `complypackDigestsByEvaluator` or at the `state.json` ingestion layer, happy to track that as a separate issue if you think it is worth pursuing.

openspec/changes/complypack-version-eviction/specs/version-eviction/spec.md
sonupreetam: complyctl currently has no mechanism to select a specific complypack version at scan time, there is no `--complypack-version` flag or config option to pin a version for comparison. `LookupByEvaluatorID` picks whichever directory `os.ReadDir` returns first, which is filesystem-dependent and non-deterministic. This is the bug filed in #645.

The "compare versions" workflow described does not exist today. Users fetch complypacks via `complyctl get` which pulls from the OCI registry tag configured in `complytime.yaml`. There is no way to request an older version or switch between versions during testing.

Your suggestion (semver sort, use highest) would fix the non-determinism (#645) but would not address #646 (unbounded disk usage from accumulating old versions). Auto-eviction solves both issues.

If version pinning or comparison becomes a requirement in the future, it would need a proper CLI flag and config option, keeping stale non-deterministic directories as an implicit mechanism for that is not the right approach. Happy to discuss tracking version pinning as a separate feature if there is a concrete use case.

sonupreetam:
@yvonnedevlinrh I applied your suggestion, would need a re-review.

yvonnedevlinrh (approved):
LGTM

cmd/complyctl/cli/cli_test.go
trevor-vaughan: A separate issue would be good. Ideally, internal consistency would be guaranteed at all interfaces.

openspec/changes/complypack-version-eviction/specs/version-eviction/spec.md
trevor-vaughan: I would go with `keep last X` (configurable) over a base eviction (maybe 3 if you don't want to add a config option).

In terms of the use case, I tried to use it this way because I wanted to compare things but the functionality was missing but I couldn't. So the mechanism is currently to download (cache) -> Try -> download (cache)  -> Manually Delete -> download(cache) ... again but why? -> Try.

trevor-vaughan: @marcusburghardt Would you mind weighing in here. I don't want to block on this but I think that it's worth a discussion before we pick a path.

marcusburghardt: Great work on this PR, @sonupreetam. The fix is well-structured, the spec artifacts are thorough, and the test coverage is solid.

Thanks @trevor-vaughan and @yvonnedevlinrh for the review. This discussion made me think deeper about the eviction vs. retention trade-off. After mapping the main use cases, such as version comparison during tests/productization, rollback after a broken publish, and cache growth over time,  I concluded this is a problem that deserves a well-designed feature rather than a quick adjustment to this PR.
I captured the use cases and design considerations that came to my mind in #676 and would be great to get more inputs and elaborate together on the right solution after the stable release.

On the other hand, the non-deterministic version selection (#645) and unbounded disk growth (#646) are real bugs that need fixing now. This PR solves them with a simple approach that doesn't change any public API, so we can evolve the cache behavior later without breaking changes.

@sonupreetam, regarding the digest validation suggestion that @trevor-vaughan raised, that's also worth filing as a separate issue when you have a chance.

WDYT?

sonupreetam: Thanks @marcusburghardt for the thorough review and for filing #676, that captures the version retention use case well. I also agreed it deserves a proper design after the stable release rather than being squeezed into this fix.

Filed #677 for the digest validation suggestion from @trevor-vaughan. Thank you @trevor-vaughan for driving this discussion.

trevor-vaughan: Approving based on the resolution of this conversation and the follow up issues

trevor-vaughan (approved):

Issue #653, opened 2026-06-30
Testing: Add container-based acceptance tests covering full user lifecycle


trevor-vaughan:
### Summary

complyctl has good coverage across unit, E2E, integration, cross-repo, and behavioral tiers, but no single test exercises the complete user journey from a blank slate through report generation. The `init` and `doctor` commands have zero E2E or integration coverage, and several recent features (`--show-passing`, complypack pull, error exit codes) lack acceptance-level verification.

This issue proposes a new acceptance test tier using `podman-compose`/`docker compose` to orchestrate a multi-service stack (real OCI registry, complyctl, test provider) that eliminates environment bleed from provider PATH discovery, `~/.complytime/` cache state, environment variables, and OCI auth credentials. The project already uses `podman` in CI (`make test-devcontainer`) and ships a `Containerfile` in `.devcontainer/`, so the infrastructure pattern is proven.

#### What exists today

| Tier | Lifecycle steps covered | `init` | `doctor` | Isolation |
|------|------------------------|--------|----------|-----------|
| Unit (34 files) | Individual functions | Wiring only | Internal logic | `t.TempDir()` |
| E2E (Go, `//go:build e2e`) | `get` -> `list` -> `generate` -> `scan` | No | No | `t.TempDir()` + in-process httptest |
| Integration (shell) | `get` -> `list` -> `generate` -> `scan` + workspace config | No | No | `mktemp -d` + standalone mock registry |
| Cross-repo (shell) | `get` -> `generate` -> `scan` with real Ampel + OPA | No | No | `mktemp -d` |
| Behavioral (Go) | Security properties across `get` -> `generate` -> `scan` | No | No | `mktemp` + in-process mock registry |

#### What is missing

- No test runs `init` -> `doctor` -> `get` -> `generate` -> `scan` -> verify reports
- `complyctl init` (`cmd/complyctl/cli/init.go`) uses interactive prompts; no test verifies the generated config is valid for downstream commands
- `complyctl doctor` has only a unit-level wiring test (`cli_test.go`) and internal package tests (`doctor_test.go`); no test exercises it as a pre-flight check in a real workflow
- `--show-passing` flag is only unit-tested in `scan_summary_test.go`
- Complypack pull/sync is absent from the standard E2E suite (only tested in cross-repo)
- Error exit codes (`ScanResponse.errors`, exit 1 on operational errors) lack dedicated acceptance coverage
[...]

PR #663, opened 2026-07-01
test(cache): fix sync test error type, add concurrency tests, remove duplicate

trevor-vaughan:
TestSync_FailureOnMissingPolicy previously passed for the wrong reason:
the mock returned a plain error instead of registry.ErrVersionNotFound,
causing it to hit the "registry unreachable" branch instead of the
"not found" branch. Fix the mock to return the correct error type.

Add TestSync_RegistryUnreachableError to cover the "registry unreachable"
branch with a non-ErrVersionNotFound error via the new SetError method.

Replace the sequential TestSync_StressConcurrentFailures with two real
concurrent tests (ConcurrentDifferentPolicies and ConcurrentMixedFailures)
that use goroutines and sync.WaitGroup to catch data races under -race.

Fix TestSync_SHA384Digest to seed the mock with the sha384 lookup ref
so sync succeeds, and assert NoError instead of silently discarding it.

Fix isDuplicateErr to use strings.Contains instead of exact match,
since OCI store errors include digest and media type context.

Remove TestComplypackSync_UnsignedWarning which was functionally identical
to TestComplypackSync_FetchAndStore and tested nothing about unsigned
warnings.

Refs: #653

Assisted-by: Claude Opus 4.6

### DIFF
--- a/internal/cache/cachetest/mock_source.go
+++ b/internal/cache/cachetest/mock_source.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"sync"
 
@@ -15,6 +16,9 @@ import (
 	"github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	ocistore "oras.land/oras-go/v2/content/oci"
+	"oras.land/oras-go/v2/errdef"
+
+	"github.com/complytime/complyctl/internal/registry"
 )
 
 // MockPolicySource provides an in-memory mock for testing sync operations.
@@ -23,6 +27,7 @@ import (
 type MockPolicySource struct {
 	mu            sync.RWMutex
 	policies      map[string]*mockPolicyData
+	errors        map[string]error
 	LastLookupRef string
 }
 
@@ -35,7 +40,18 @@ type mockPolicyData struct {
 func NewMockPolicySource() *MockPolicySource {
 	return &MockPolicySource{
 		policies: make(map[string]*mockPolicyData),
+		errors:   make(map[string]error),
 	}
+}
+
+// SetError injects a custom error for a specific policy lookup. When set,
+// DefinitionVersion returns this error instead of performing the normal
+// lookup. This allows tests to simulate arbitrary registry failures
+// (e.g., network errors) distinct from the "not found" path.
+func (m *MockPolicySource) SetError(policyID string, err error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.errors[policyID] = err
 }
 
 // SeedPolicy adds a mock policy for testing. The policy is registered under
@@ -54,6 +70,9 @@ func (m *MockPolicySource) SeedPolicy(policyID, version, digestStr string) {
 
 // DefinitionVersion returns digest and version for a mock policy.
 // The lookupRef is recorded in LastLookupRef for test assertions.
+// If SetError was called for the lookupRef, the injected error is
+// returned. If the policy is not seeded, registry.ErrVersionNotFound
+// is returned, matching real registry behavior.
 func (m *MockPolicySource) DefinitionVersion(_ context.Context, lookupRef string) (string, string, error) {
 	m.mu.Lock()
 	m.LastLookupRef = lookupRef
@@ -61,9 +80,14 @@ func (m *MockPolicySource) DefinitionVersion(_ context.Context, lookupRef string
 
 	m.mu.RLock()
 	defer m.mu.RUnlock()
+
+	if injected, ok := m.errors[lookupRef]; ok {
+		return "", "", injected
+	}
+
 	p, ok := m.policies[lookupRef]
 	if !ok {
-		return "", "", fmt.Errorf("policy %s not found", lookupRef)
+		return "", "", fmt.Errorf("%w", registry.ErrVersionNotFound)
 	}
 	return p.digest, p.version, nil
 }
@@ -75,7 +99,7 @@ func (m *MockPolicySource) CopyPolicy(ctx context.Context, policyID, tag string,
 	_, ok := m.policies[policyID]
 	m.mu.RUnlock()
 	if !ok {
-		return ocispec.Descriptor{}, fmt.Errorf("policy %s not found in mock source", policyID)
+		return ocispec.Descriptor{}, fmt.Errorf("policy %s: %w", policyID, registry.ErrVersionNotFound)
 	}
 
 	configData := []byte("{}")
@@ -195,8 +219,5 @@ func (m *MockBundlePolicySource) CopyPolicy(ctx context.Context, policyID, tag s
 }
 
 func isDuplicateErr(err error) bool {
-	if err == nil {
-		return false
-	}
-	return err.Error() == "already exists"
+	return errors.Is(err, errdef.ErrAlreadyExists)
 }
--- a/internal/cache/cachetest/mock_source.go
+++ b/internal/cache/cachetest/mock_source.go
@@ -87,7 +87,7 @@ func (m *MockPolicySource) DefinitionVersion(_ context.Context, lookupRef string
 
 	p, ok := m.policies[lookupRef]
 	if !ok {
-		return "", "", fmt.Errorf("%w", registry.ErrVersionNotFound)
+		return "", "", fmt.Errorf("policy %s: %w", lookupRef, registry.ErrVersionNotFound)
 	}
 	return p.digest, p.version, nil
 }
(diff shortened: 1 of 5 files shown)


internal/cache/sync_test.go
```diff
@@ -187,67 +218,133 @@ func TestSync_RedownloadAfterDeletion(t *testing.T) {
...
+	var wg gosync.WaitGroup
+	wg.Add(workers)
 
-	for i := 0; i < iterations; i++ {
-		state, loadErr := cache.LoadState(cacheDir)
-		require.NoError(t, loadErr, "state load must not fail on iteration %d", i)
+	for i := 0; i < workers; i++ {
+		go func(workerID int) {
+			defer wg.Done()
+			state, loadErr := cache.LoadState(cacheDir)
```
marcusburghardt: **[HIGH] Race condition on `state.json` causes CI failure**

All 10 goroutines share the same `cacheDir` and each calls `cache.LoadState(cacheDir)` inside the goroutine. When `SyncPolicy` completes, it calls `SaveState()` → `os.WriteFile(statePath, data, 0600)`, which truncates the file before writing. A concurrent `LoadState` from another goroutine observes the truncated file and fails with `unexpected end of JSON input`.

CI log: `worker 6: state load failed: failed to parse state file .../state.json: unexpected end of JSON input`

`TestSync_ConcurrentMixedFailures` has the same structural pattern but is less likely to trigger because odd-numbered workers fail at `DefinitionVersion` (before `SaveState`), reducing concurrent writers from 10 to ~5.

Suggested fixes (any one resolves the failure):

1. **Per-worker cache directories** (simplest): give each goroutine its own `cacheDir` subdirectory. Tests concurrent mock access without file-level contention.
2. **Pre-load shared state, serialize saves**: load state once before spawning, share `*State`, add a mutex around `UpdatePolicyState` + `SaveState`.
3. **Atomic writes in `SaveState`** (production code change): write-then-rename pattern. Fixes the root cause but changes production scope — likely a separate PR per Constitution III.

marcusburghardt (changes requested):
Test quality improvements for `internal/cache/` — fixes mock error type, adds concurrency tests, removes duplicate test. The `isDuplicateErr` fix, `TestSync_RegistryUnreachableError` addition, SHA384 test fix, and duplicate removal are all correct and improve test fidelity.

However, `TestSync_ConcurrentDifferentPolicies` has a race condition on `state.json` file I/O that causes the `unit-test` CI check to fail. See inline comment for details and suggested fixes.

marcusburghardt (dismissed):
LGTM

internal/cache/cachetest/mock_source.go
```diff
@@ -198,5 +221,5 @@ func isDuplicateErr(err error) bool {
 	if err == nil {
 		return false
 	}
-	return err.Error() == "already exists"
+	return strings.Contains(err.Error(), "already exists")
```
sonupreetam: **[HIGH] Use `errors.Is` instead of `strings.Contains` for error matching (CS-006)**

The old exact match never worked because `ocistore.Store.Push()` wraps `errdef.ErrAlreadyExists` with digest and media type context (e.g., `"sha256:abc...: application/vnd.oci.empty.v1+json: already exists"`). `strings.Contains` fixes the immediate problem but is fragile -- it relies on the error message text remaining stable across ORAS versions.

Every other caller in the dependency tree (6 sites across oras-go, go-gemara, complypack) uses `errors.Is`:

```suggestion
	return errors.Is(err, errdef.ErrAlreadyExists)
```

Requires adding `"errors"` to the import block (it's already imported as `strings` is) and `errdef "oras.land/oras-go/v2/errdef"` to the import section.

internal/cache/cachetest/mock_source.go
```diff
@@ -54,16 +69,24 @@ func (m *MockPolicySource) SeedPolicy(policyID, version, digestStr string) {
...
 func (m *MockPolicySource) DefinitionVersion(_ context.Context, lookupRef string) (string, string, error) {
 	m.mu.Lock()
 	m.LastLookupRef = lookupRef
 	m.mu.Unlock()
 
 	m.mu.RLock()
 	defer m.mu.RUnlock()
+
+	if injected, ok := m.errors[lookupRef]; ok {
+		return "", "", injected
```
sonupreetam, line 85: **[MEDIUM] Mock error format diverges from production, causing double-wrapped "policy" prefix**

The production `DefinitionVersion` (via `fetchVersion` in `registry/client.go:138`) returns `fmt.Errorf("%w: %s tag %q", ErrVersionNotFound, ref, tag)` -- the sentinel first, then context. `SyncPolicy` then wraps with `fmt.Errorf("policy %s: %w", policyID, err)`.

This mock wraps as `fmt.Errorf("policy %s: %w", lookupRef, ErrVersionNotFound)`, so `SyncPolicy` produces `"policy X: policy X: version not found in registry"` -- redundant prefix.

Consider matching the production format or simplifying to just `fmt.Errorf("%w", registry.ErrVersionNotFound)`, which keeps mock errors clean and avoids simulating formatting the mock doesn't need to test.

sonupreetam (changes requested):
Good fix -- `TestSync_FailureOnMissingPolicy` was indeed exercising the wrong error branch. The `ErrVersionNotFound` wrapping, `SetError` addition, concurrency test rewrite, and SHA384 seeding fix are all solid improvements.

Two items on code introduced in this PR:

1. `isDuplicateErr` should use `errors.Is(err, errdef.ErrAlreadyExists)` instead of `strings.Contains` -- matches every other caller in the dependency tree and is immune to message format changes.

2. The mock's error format `fmt.Errorf("policy %s: %%w", ...)` diverges from production, causing `SyncPolicy` to produce a double-wrapped `"policy X: policy X: ..."` message.

sonupreetam (dismissed):
LGTM

marcusburghardt (dismissed):
LGTM

marcusburghardt:
@trevor-vaughan there were two PRs editing the `.gitattributes` file slightly different. The merged one is cleaner so we can remove the `.gitattributes` file here to resolve the conflict.

trevor-vaughan:
> @trevor-vaughan there were two PRs editing the `.gitattributes` file slightly different. The merged one is cleaner so we can remove the `.gitattributes` file here to resolve the conflict.

The merged one will result in not recursing and having issues. This one fixes that.

marcusburghardt:
> > @trevor-vaughan there were two PRs editing the `.gitattributes` file slightly different. The merged one is cleaner so we can remove the `.gitattributes` file here to resolve the conflict.
> 
> The merged one will result in not recursing and having issues. This one fixes that.

Unfortunately I this is not what the official documentation says and can be confirmed with simple tests.
Both patterns are functionally identical in `.gitattributes`. Git's pattern matching follows `gitignore(5)` rules (per [`gitattributes(5)`](https://git-scm.com/docs/gitattributes)):

> The rules by which the pattern matches paths are the same as in `.gitignore` files

And [`gitignore(5)`](https://git-scm.com/docs/gitignore#_pattern_format) defines:

> A trailing `"/**"` matches everything inside. For example, `"abc/**"` matches all files inside directory `"abc"`, relative to the location of the `.gitignore` file, **with infinite depth**.

### Reproduction

```bash
cd /tmp && rm -rf gitattr-proof && mkdir gitattr-proof && cd gitattr-proof && git init -q
mkdir -p vendor/a/b/c
echo x > vendor/top.go && echo x > vendor/a/mid.go && echo x > vendor/a/b/c/deep.go
printf '* text eol=lf\nvendor/** -text\n' > .gitattributes && git add .
echo "--- vendor/** ---" && git check-attr text -- vendor/top.go vendor/a/mid.go vendor/a/b/c/deep.go
sed -i 's|vendor/\*\*|vendor/**/*|' .gitattributes && git add .gitattributes
echo "--- vendor/**/* ---" && git check-attr text -- vendor/top.go vendor/a/mid.go vendor/a/b/c/deep.go
rm -rf /tmp/gitattr-proof
```

Output is identical for both patterns at all nesting levels. Now that this is clarified, lets update the PRs and move forward. Thanks!

marcusburghardt (changes requested):
https://github.com/complytime/complyctl/pull/663#issuecomment-4923165138

trevor-vaughan:
@marcusburghardt You're absolutely right about the path matching. I was misremembering :face_exhaling:.

It turns out that we both missed something that I hadn't been aware of. Apparently, the best form to use is `vendor/** linguist-generated -diff -text` which will correct it both in GitHub scans and locally. Fixed in the rebased update.

sonupreetam (approved):
LGTM

marcusburghardt:
> @marcusburghardt You're absolutely right about the path matching. I was misremembering 😮‍💨.
> 
> It turns out that we both missed something that I hadn't been aware of. Apparently, the best form to use is `vendor/** linguist-generated -diff -text` which will correct it both in GitHub scans and locally. Fixed in the rebased update.

I see @trevor-vaughan . I was even checking this morning the documentation about `linguist-generated` and `linguist-vendored` which were also both unfamiliar to me. It seems the `linguist-vendored` would fit better in this case but it would not hide vendored files in diff, which seems the main delta with `linguist-generated`. Considering we have other guardrails such as OSV-Scanner, Trivy, CodeQL we should be fine to hide the diff.

As a retrospective, the main focus of this PR were the tests and not the `.gitattributes`. Things were improved but it would be great if we could agree to separate these changes and keep the PRs focused so we can avoid similar situations and merge the PRs more quickly with less friction. Are we good? : )

marcusburghardt (approved):
LGTM

marcusburghardt:
Just rebased using the web interface. This will preserve the previous approvals.

PR #665, opened 2026-07-01
test(output): add NeedsReview, execution plan, OSCAL, and content tests

trevor-vaughan:
Add execution_plan_test.go with 7 tests for FormatExecutionPlan and
FormatPreScanSummary. Add NeedsReview result tests to scan_summary_test.go
and markdown_test.go. Enhance TestEvaluator_Write to verify file content
fields instead of just checking file existence.

Fix trivially true assertions in scan summary tests: replace Contains
checks that matched any output (like checking for "-") with targeted
assertions verifying the correct fallback values in table positions.

Add positive assertion for empty step identity instead of relying only
on NotContains.

Add OSCAL result mapping tests for Failed, Unknown, and default (NotRun,
NeedsReview) branches -- previously only 2 of 5 branches in
gemaraResultToOSCAL had coverage.

Refs: #653

Assisted-by: Claude Opus 4.6

### DIFF
--- a/internal/output/execution_plan_test.go
+++ b/internal/output/execution_plan_test.go
@@ -1,0 +1,107 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package output
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/complytime/complyctl/internal/complytime"
+)
+
+func TestFormatExecutionPlan_SingleRow(t *testing.T) {
+	rows := []ExecutionPlanRow{
+		{
+			TargetID:         "my-app",
+			ProviderID:       "ampel",
+			RequirementCount: 5,
+			Status:           "healthy",
+		},
+	}
+
+	result := FormatExecutionPlan("test-policy", rows)
+
+	assert.Contains(t, result, "Execution Plan: test-policy")
+	assert.Contains(t, result, "Target: my-app")
+	assert.Contains(t, result, "Provider: ampel")
+	assert.Contains(t, result, complytime.StatusPassed)
+	assert.Contains(t, result, "healthy")
+	assert.Contains(t, result, "Requirements: 5")
+	assert.Contains(t, result, "Generation completed.")
+}
+
+func TestFormatExecutionPlan_MultipleRows(t *testing.T) {
+	rows := []ExecutionPlanRow{
+		{
+			TargetID:         "app-1",
+			ProviderID:       "ampel",
+			RequirementCount: 3,
+			Status:           "healthy",
+		},
+		{
+			TargetID:         "app-2",
+			ProviderID:       "opa",
+			RequirementCount: 7,
+			Status:           "unavailable",
+		},
+	}
+
+	result := FormatExecutionPlan("multi-policy", rows)
+
+	assert.Contains(t, result, "Execution Plan: multi-policy")
+	assert.Contains(t, result, "Target: app-1")
+	assert.Contains(t, result, "Provider: ampel")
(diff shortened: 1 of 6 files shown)


internal/output/scan_summary_test.go
```diff
@@ -383,6 +399,65 @@ func TestFormatScanSummary_ShowPassingFalse_AllPassed_NoTable(t *testing.T) {
 	assert.Contains(t, output, "2 requirements: 2 passed, 0 failed, 0 not applicable, 0 skipped, 0 errors")
```
gvauter, line 399: **[MEDIUM] Misleading test name (TC-007)**: This test is named `NeedsReviewResult` but the test body uses `provider.ResultError`, which maps to `gemara.Unknown`, not `gemara.NeedsReview`. The 15-line comment block acknowledges this ("To truly hit NeedsReview, we'd need gemara internals"). The assertions check for `StatusError` and `"1 errors"` -- which is Unknown/error behavior, not NeedsReview-specific.

`TestSortPriority_NeedsReview` below does cover the sort priority branch directly, so the NeedsReview *sort ordering* is tested. But this test's name could mislead future maintainers into thinking NeedsReview has integration-level scan summary coverage.

Consider renaming to `TestFormatScanSummary_ErrorResultDefaultBranch` or similar to match what it actually exercises.

gvauter (dismissed):
Clean, well-scoped test improvement PR. Adds meaningful coverage for NeedsReview (markdown, OSCAL, sort priority), execution plan formatting, and OSCAL result mapping branches. Fixes trivially-true assertions in scan summary tests. One MEDIUM nit on a misleading test name -- non-blocking.

internal/output/execution_plan_test.go
```diff
@@ -0,0 +1,109 @@
...
+	result := FormatPreScanSummary(5, []string{"ampel"}, []string{"my-target"})
+
+	assert.Equal(t, "Scanning 5 requirements via ampel for target(s): my-target...", result)
+}
+
+func TestFormatPreScanSummary_EmptyInputs(t *testing.T) {
+	result := FormatPreScanSummary(0, nil, nil)
+
+	assert.Contains(t, result, "Scanning 0 requirements")
+	assert.Contains(t, result, "via ")
```
yvonnedevlinrh: LOW The assertions Contains(result, "via ") and Contains(result, "target(s): ") match the format string template itself, so they pass regardless of whether the empty slices were handled correctly. A more precise check would be an exact string match: assert.Equal(t, "Scanning 0 requirements via  for target(s): ...", result) or checking for the double-space that indicates an empty strings.Join: assert.Contains(t, result, "via  for")

This review was generated by /review-pr (AI-assisted).

yvonnedevlinrh (dismissed):
The coverage additions for FormatExecutionPlan, FormatPreScanSummary, gemaraResultToOSCAL (all 6 branches), and NeedsReview across markdown/scan-summary are well-targeted. The fix of trivially-true assertions (e.g., replacing Contains(output, "-") with field-position checks) is particularly valuable -- those were giving false confidence.

One minor note: TestFormatPreScanSummary_EmptyInputs (execution_plan_test.go:105-109) uses Contains(result, "via ") and Contains(result, "target(s): ") which match the format template itself and will pass regardless of how empty slices are handled. An exact match or checking for the double-space from an empty join would be more precise. Non-blocking.

George's naming comment on TestFormatScanSummary_NeedsReviewResult is worth addressing -- agree that the test name should reflect what it actually exercises (the default/error branch).

marcusburghardt (approved):
LGTM

yvonnedevlinrh (approved):
LGTM

PR #666, opened 2026-07-01
test(policy): add happy-path loader tests and content assertions

trevor-vaughan:
ResolveVersion's success paths (direct tag match and fallback to last
cached version) previously had zero happy-path coverage. Add tests for
both paths. Add tests for ListCachedPolicies which had zero coverage.

Verify parsed content (Controls and Guidelines fields) in the split-layer
resolver test to match the symmetric coverage already present in the
bundle-path equivalent test.

Refs: #653

Assisted-by: Claude Opus 4.6

### DIFF
--- a/internal/policy/loader_test.go
+++ b/internal/policy/loader_test.go
@@ -66,6 +66,58 @@ func TestGetCachedVersions_NonexistentPolicy(t *testing.T) {
 	assert.Empty(t, versions)
 }
 
+// --- ResolveVersion happy-path tests ---
+
+func TestResolveVersion_DirectTagMatch(t *testing.T) {
+	loader := seedTestPolicy(t, "resolve-direct", "v1.0.0")
+
+	version, err := loader.ResolveVersion("resolve-direct", "v1.0.0")
+	require.NoError(t, err)
+	assert.Equal(t, "v1.0.0", version)
+}
+
+func TestResolveVersion_LatestFallbackToCachedVersion(t *testing.T) {
+	loader := seedTestPolicy(t, "resolve-latest", "v1.0.0")
+
+	version, err := loader.ResolveVersion("resolve-latest", "")
+	require.NoError(t, err)
+	assert.Equal(t, "v1.0.0", version, "should fall back to the last cached version when 'latest' tag doesn't exist")
+}
+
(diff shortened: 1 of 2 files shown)


internal/policy/loader_test.go
```diff
@@ -66,6 +66,58 @@ func TestGetCachedVersions_NonexistentPolicy(t *testing.T) {
...
+	require.NoError(t, err)
+
+	loader := NewLoader(cacheMgr)
+	result, err := loader.ListCachedPolicies()
+	require.NoError(t, err)
+	require.Contains(t, result, "policy-a")
+	assert.Equal(t, []string{"v1.0.0"}, result["policy-a"])
+}
+
 // --- T156: LoadLayerByMediaType ---
```
marcusburghardt, line 121: **[MEDIUM] DRY: duplicates `seedTestPolicy` helper (CS-004)**

This function manually creates mock source, cache manager, state, and sync — all of which `seedTestPolicy` (line 112-129) already does. The only difference is the digest string, which is immaterial to `ListCachedPolicies`.

```suggestion
func TestListCachedPolicies_SeededCache(t *testing.T) {
	loader := seedTestPolicy(t, "policy-a", "v1.0.0")

	result, err := loader.ListCachedPolicies()
	require.NoError(t, err)
	require.Contains(t, result, "policy-a")
	assert.Equal(t, []string{"v1.0.0"}, result["policy-a"])
}
```

marcusburghardt (approved):
Well-scoped test-only PR that fills genuine coverage gaps. `ResolveVersion` had zero happy-path coverage for both its direct-tag-match and latest-fallback paths; `ListCachedPolicies` had zero coverage entirely. The resolver content assertions establish symmetry with the bundle-path test. One minor DRY suggestion below.

hbraswelrh (approved):
LGTM.

internal/policy/loader_test.go
marcusburghardt: @trevor-vaughan up to you if you prefer to accept this suggestion or merge as it is. :) Only let us know if we need to approve again if changed. Thanks.

marcusburghardt: There is already two approvals. I will merge it now and create a follow-up PR.

PR #700, opened 2026-07-07
fix: use config.Version for complypack state to match on-disk directory

### LINKED ISSUE #694: bug: complypack state version vs on-disk directory name mismatch
**Describe the bug**
State records complypack versions using the OCI tag string (e.g., `v1.0.0`), but `Store()` creates on-disk directories using the version from `complypack.Config` (e.g., `1.0.0` without the `v` prefix). This causes a mismatch between state tracking and the filesystem layout, resulting in false "orphaned" warnings from `complyctl doctor` and broken state-driven lookups.

**To Reproduce**
1. Configure complypacks in `complytime.yaml` with a `v`-prefixed tag (e.g., `v1.0.0`)
2. Run `complyctl get` to fetch complypacks
3. Inspect state: `cat ~/.complytime/state.json | grep version` — shows `"version": "v1.0.0"`
4. Inspect disk: `ls ~/.complytime/complypacks/ampel/` — shows `1.0.0` (no `v` prefix)
5. Run `complyctl doctor`
6. See: `orphaned complypack version 1.0.0 for evaluator ampel — not referenced in state.json`

**Expected behavior**
`complyctl doctor` should not report freshly cached complypacks as orphaned. The version in state.json and the on-disk directory name should be consistent, or lookups should normalize between the two conventions.

**Screenshots**
N/A

**Additional context**
Root cause: `SyncComplypack` calls `UpdateComplypackStateWithVerification(repository, version, ...)` where `version` is the OCI tag (`v1.0.0`). But `Store()` uses `config.Version` from the complypack's internal `config.json`, which uses bare semver (`1.0.0`). The two conventions are never reconciled.

This also affects `LookupByEvaluatorID` (state-driven lookup fails, falls back to directory scan), `buildTrackedVersionSet` (eviction misclassifies tracked versions as orphaned), and `tryLocalCacheHit` (local cache hit lookup fails by version).

Suggested fix: normalize the version in `UpdateComplypackStateWithVerification` to use the version from `complypack.Config` (matching the on-disk directory name), or add normalization at comparison points.

Discovered during #676 implementation.

yvonnedevlinrh:
## Summary

Fixes #694

State records complypack versions using the OCI tag string (e.g., `v1.0.0`), but `Store()` creates on-disk directories using the version from `complypack.Config` (e.g., `1.0.0` without the `v` prefix). This mismatch causes state-driven lookups to fail and downstream consumers to misclassify valid cache entries.

### Fix (`internal/cache/complypack_sync.go`)
- `SyncComplypack()`: use `result.Config.Version` (from the complypack's embedded `config.json`) instead of the OCI tag `version` when calling `UpdateComplypackStateWithVerification()`, so the state version matches the on-disk directory name

### Tests (`internal/cache/complypack_sync_test.go`)
- Add `seedComplypackWithTag()` helper that allows seeding a mock complypack where the OCI tag differs from `config.Version` (e.g., tag `v1.0.0` vs config `1.0.0`)
- Add `TestComplypackSync_VPrefixedTag_StateMatchesDisk` regression test verifying:
  - State records `"1.0.0"` (config.Version), not `"v1.0.0"` (OCI tag)
  - On-disk directory is `1.0.0/`, not `v1.0.0/`
  - No directory exists at the OCI tag path
  - Incremental sync skip works correctly after the fix

## Root Cause

In `SyncComplypack()`, two different version values were used for the same artifact:
- **Line 134**: `Store(result.Config, result.Content)` creates the on-disk directory using `result.Config.Version` (e.g., `"1.0.0"`)
- **Line 139**: `UpdateComplypackStateWithVerification(repository, version, ...)` records the OCI tag (e.g., `"v1.0.0"`)

These two values were never reconciled. Existing tests didn't catch this because the mock's `seedComplypack()` uses a single `version` parameter for both the OCI tag and config version, making the divergence scenario untestable.

## Impact

- State and filesystem are now consistent for complypack cache entries
- Eliminates false state/disk mismatches that would affect downstream features relying on state-driven version lookups (eviction, doctor orphan checks, local cache hits)
- No functional behavior changes for users whose OCI tags already match their complypack config versions

## Review Hints

```shell
# Run the regression test
go test -race -v -run "TestComplypackSync_VPrefixedTag" ./internal/cache/

# Run all complypack sync tests
go test -race -v -run "TestComplypackSync" ./internal/cache/

# Lint
make lint
```

### DIFF
--- a/internal/cache/complypack_sync_test.go
+++ b/internal/cache/complypack_sync_test.go
@@ -59,6 +59,25 @@ func (m *mockComplypackSource) seedComplypack(repository, evaluatorID, version, 
 	m.packs[repository+":"+version] = data
 }
 
+// seedComplypackWithTag registers a complypack where the OCI tag differs from
+// the config version (e.g., tag "v1.0.0" vs config version "1.0.0"). The
+// DefinitionVersion lookup uses the tag for the versioned key, while
+// CopyComplypack packs the artifact with the config version.
+func (m *mockComplypackSource) seedComplypackWithTag(
+	repository, evaluatorID, tag, configVersion, digestStr, content string,
+) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	data := &mockComplypackData{
+		digest:      digestStr,
+		version:     configVersion,
+		evaluatorID: evaluatorID,
+		content:     content,
+	}
+	m.packs[repository] = data
+	m.packs[repository+":"+tag] = data
+}
+
 func (m *mockComplypackSource) DefinitionVersion(_ context.Context, lookupRef string) (string, string, error) {
 	m.mu.RLock()
 	defer m.mu.RUnlock()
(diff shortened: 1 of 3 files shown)


marcusburghardt (approved):
LGTM

gvauter (dismissed):
LGTM

internal/cache/complypack_sync_test.go
```diff
@@ -59,6 +59,25 @@ func (m *mockComplypackSource) seedComplypack(repository, evaluatorID, version,
...
+	repository, evaluatorID, tag, configVersion, digestStr, content string,
+) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	data := &mockComplypackData{
+		digest:      digestStr,
+		version:     configVersion,
+		evaluatorID: evaluatorID,
+		content:     content,
+	}
```
marcusburghardt, line 77: **[CRITICAL] `seedComplypackWithTag` missing `id` field — causes `TestComplypackSync_VPrefixedTag_StateMatchesDisk` to fail**

The existing `seedComplypack` helper sets `id: "test-" + evaluatorID` (line 55), but `seedComplypackWithTag` omits this field. When `CopyComplypack` builds `complypack.Config{ID: p.id, ...}`, the empty `ID` causes `complypack.Pack()` to reject the config with `invalid config: id is required`.

```suggestion
	data := &mockComplypackData{
		digest:      digestStr,
		version:     configVersion,
		evaluatorID: evaluatorID,
		id:          "test-" + evaluatorID,
		content:     content,
	}
```

marcusburghardt (changes requested):
Production fix is correct and well-scoped — using `result.Config.Version` instead of the OCI tag `version` directly addresses the state/filesystem mismatch described in #694. The regression test structure is thorough (verifies state version, on-disk directory, absence of v-prefixed directory, incremental skip), but has a data setup bug that prevents it from passing.

Pre-existing Packit/COPR infrastructure failures (rpm-build, testing-farm) are unrelated to this PR.

marcusburghardt:
packit failures are caused by a typo in RPM spec file. This PR fix it: https://github.com/complytime/complyctl/pull/703

internal/cache/complypack_sync_test.go
yvonnedevlinrh: Addressed in 5f16ac7d: added missing id: "test-" + evaluatorID field to seedComplypackWithTag, matching the pattern in seedComplypack. TestComplypackSync_VPrefixedTag_StateMatchesDisk now passes.

yvonnedevlinrh:
Thanks @marcusburghardt  Addressed in 5f16ac7d: added missing id: "test-" + evaluatorID field to seedComplypackWithTag, matching the pattern in seedComplypack. TestComplypackSync_VPrefixedTag_StateMatchesDisk now passes.

marcusburghardt (approved):
LGTM

trevor-vaughan (approved):

PR #702, opened 2026-07-07
feat: complypack cache version management

### LINKED ISSUE #676: enhancement: complypack cache version management — avoid re-downloads and support version switching
## Problem

The complypack cache (`~/.complytime/complypacks/{evaluator-id}/{version}/`) is a global, application-level store shared across all workspaces. It currently has no version management capabilities:

1. **`LookupByEvaluatorID` is non-deterministic** (#645): When multiple version directories exist for the same evaluator-id, the scan pipeline picks whichever `os.ReadDir` returns first. This was addressed in #661 by evicting old versions on store, ensuring only one version exists.

2. **Unbounded disk growth** (#646): When a complypack version changes between releases, `Store()` only writes the new version directory. Prior to #661, old version directories were never removed, accumulating indefinitely in `~/.complytime/complypacks/`. #661 addressed this by evicting old versions on store, but this prevents retaining previously cached versions for switching or rollback.

3. **Version switches force re-downloads**: When a user changes the complypack tag in `complytime.yaml` (e.g., from `:v2.0.0` to `:v1.0.0`), `SyncComplypack` compares the state.json digest against the remote digest, finds a mismatch, and re-fetches from the registry — even if the target version was previously cached locally.

4. **No mechanism to switch versions**: There is no CLI flag or config option to select a specific cached complypack version. The cache is opaque to the user.

Originated from [this review discussion](https://github.com/complytime/complyctl/pull/661#discussion_r3507023156) on #661.

### UC-1: Productization — comparing policy and complypack combinations

Multiple teams release policies and complypacks independently. During productization, QA engineers need to analyze delta results across different combinations to confirm expected changes:

```
Config → policy-A:v1 + complypack-X:v1 → scan → results-baseline
Config → policy-A:v2 + complypack-X:v1 → scan → results-candidate
Compare results
```

Today, switching between versions in `complytime.yaml` triggers a re-download each time, even when both versions were recently fetched. With a shared global cache, workspace isolation does not help — all workspaces hit the same cache.

### UC-2: Rollback after a broken publish

A team accidentally publishes a broken complypack. The user notices incorrect scan results and wants to revert to the previous version:

```
complyctl get          → fetches broken v2.0.0
(user notices bad results)
Config → :v1.0.0
complyctl get          → must re-download v1.0.0 from registry
```

If the registry is temporarily unavailable or slow, the rollback is blocked. The previous version was on disk moments ago but was evicted.

### UC-3: Normal operation — always use newest

End users typically want the newest working version. They use multiple policies but not multiple versions of the same policy. Similarly for complypacks — the newest version is expected to be the best one. The cache should work transparently without manual management for this majority case.

### UC-4: Long-running installations — unnoticed cache growth

A user runs complyctl over months or years, syncing complypack updates as teams publish new versions. The user never runs a cleanup command because they are unaware the cache grows. Over time, stale versions accumulate in `~/.complytime/complypacks/` consuming disk space with no visibility or feedback.

The challenge: without a user-defined retention policy (e.g., "keep last N versions" or "max 500 MB"), there is no obvious threshold to determine when the cache should be cleaned. The system needs a way to surface cache health to the user without requiring proactive action.

This concern is amplified in environments that perform compatibility matrix validation (see [complytime/complytime#20](https://github.com/complytime/complytime/pull/20)), where teams cycle through multiple version combinations across validation rounds, accumulating cached artifacts from each tested combination.

## Current State

| Aspect | Behavior |
|--------|----------|
| Cache structure | `{evaluator-id}/{version}/` — already supports multiple versions on disk |
| Version selection | `LookupByEvaluatorID` — blind directory scan, no state awareness |
| Sync check | state.json digest vs remote digest only — no local cache existence check |
| Disk cleanup | None (before #661) / auto-eviction on store (after #661) |
| User control | None — no version listing, selection, or cleanup command |

## Requirements

| # | Requirement | Use case |
|---|-------------|----------|
| R-1 | Deterministic version selection based on what was last synced | UC-1, UC-2, UC-3 |
| R-2 | Avoid re-downloading OCI artifacts when switching to a previously cached version | UC-1, UC-2 |
| R-3 | Simple mechanism to control disk usage (cleanup of unused versions) | UC-3, UC-4 |
| R-4 | Proactive feedback when cached versions are stale or cache size warrants cleanup | UC-4 |

## Design Considerations

The following observations may help inform the design. They are starting points, not prescriptive solutions.

- **state.json already tracks evaluator-id and version per complypack entry** (`internal/cache/state.go:24-28`). A reverse lookup from evaluator-id to version would make `LookupByEvaluatorID` deterministic without eviction.

- **A version-specific `Lookup(evaluatorID, version)` method already exists** (`internal/cache/complypack.go:149`). The scan pipeline currently uses the version-agnostic `LookupByEvaluatorID` because it only has the evaluator-id from the policy graph — but the version could be resolved from state.

- **`SyncComplypack` could check local cache before fetching** (`internal/cache/complypack_sync.go:72-91`). If the target evaluator-id + version directory exists on disk, state can be updated without a network round-trip.

- **Version switching UX already exists at the config level**: users change the OCI tag in `complytime.yaml` and run `complyctl get`. No new flags are strictly required for switching — the improvement is making the cache serve locally cached versions.

- **`complyctl doctor` already checks complypack cache health** (`internal/doctor/doctor.go:770`). It could be extended to report stale versions or cache size, giving users visibility without requiring a separate monitoring habit. The open design question is what heuristic to use: number of unreferenced versions, total cache size, age of unreferenced entries, or a combination.

- **OCI digest comparison operates at the manifest level**, not content level. A version bump in the complypack config (e.g., `1.0.0` → `1.0.1`) produces a different manifest digest even when the functional content (policies, rules) is unchanged. This means version bumps without content changes still trigger re-fetches and create separate cache entries. This is standard OCI behavior and not something the cache layer should attempt to optimize — content-level deduplication would add complexity disproportionate to the benefit.

## Related

- #645 — `LookupByEvaluatorID` selects arbitrary version (fixed by #661)
- #646 — old complypack versions not evicted from cache (fixed by #661)
- #661 — complypack cache reliability: version eviction and generation state accuracy
- complytime/complytime#20 — ecosystem compatibility matrix (defines validated version combinations that drive cache accumulation patterns)

yvonnedevlinrh:
## Summary

Implements complypack cache version management: retention-count-aware eviction, local cache hit optimization, state-driven lookup, and doctor cache health reporting.

The complypack cache previously enforced a single-version-per-evaluator invariant, forcing re-downloads when switching between previously cached versions. This change makes the retention count configurable via `COMPLYTIME_CACHE_VERSIONS` (default 1, preserving current behavior) and adds a local cache check in `SyncComplypack` to avoid redundant network fetches.

### Changes

**Configuration**
- `COMPLYTIME_CACHE_VERSIONS` env var configures retention count per evaluator-id (default 1)
- `CacheRetentionCount()` in `internal/cache/retention.go` with validation, clamping, and warnings

**Cache eviction** (`internal/cache/complypack.go`)
- `NewComplypackCache()` gains `*State` parameter for state-driven lookup and eviction ordering
- `evictOldVersions()` retains N versions per evaluator-id, evicting orphaned dirs first then oldest by `LastUpdated` timestamp
- `LookupByEvaluatorID()` resolves from state.json with directory-scan fallback

**Sync** (`internal/cache/complypack_sync.go`)
- `SyncComplypack()` checks local cache before remote fetch
- Re-verifies via registry API when verifier is configured (FR-010)
- Returns `(true, nil)` for cache hits to trigger generation invalidation (FR-009)

**State** (`internal/cache/state.go`)
- `EvaluatorIDToVersion()` reverse lookup on `*State`

**Doctor** (`internal/doctor/doctor.go`)
- `CheckComplypacks()` reports cache disk usage and orphaned/untracked versions

## Related Issues

- Closes #676

## Review Hints

- The OpenSpec is at `openspec/changes/complypack-cache-versioning/` with proposal, design (D1-D6), spec (FR-001 to FR-010), and tasks. Reading `design.md` gives the rationale for key decisions.

- All code is under `internal/` — no public API changes, no downstream repo impact.

- `NewComplypackCache` call sites in `providers.go` and `doctor.go` pass `nil` state (Guard/Architect finding — acceptable since the directory-scan fallback works correctly; state could be loaded here in a follow-up).

- `complyctl doctor` will report orphan warnings for freshly cached complypacks due to a **pre-existing** version prefix mismatch between state.json (`v1.0.0`) and on-disk directories (`1.0.0`). This is tracked in a separate bug issue and is not introduced by this PR.

- Spec review passed (2 iterations, 6 Divisor agents). Code review passed (6 Divisor agents, 2 fixes applied).

- To run tests:

```bash
# Unit tests (all modified packages)
go test -race -count=1 ./internal/cache/... ./internal/doctor/... ./internal/complytime/... ./cmd/complyctl/cli/...

# Full CI suite
make lint && go test -race -count=1 ./...

### CODE REVIEW

**Review by gvauter — DISMISSED** [review]
## APPROVE

All 10 functional requirements (FR-001–FR-010) from the OpenSpec are implemented and tested. CI checks all pass.

**Strengths:**
- Strong spec-to-code traceability — every FR maps to a named function and dedicated test
- Thorough edge case coverage: nil state, corrupt state, empty state, cross-evaluator isolation, corrupted cache content, verifier integration on cache hits
- Graceful degradation at every callsite — nil state falls back to directory-scan behavior, so this feature cannot break existing workflows
- Default retention count of 1 preserves backward compatibility (Constitution VII)
- `nil` state in `providers.go` is an accepted trade-off documented in the PR description

**Issue #676 requirement coverage:** R-1 (deterministic selection), R-2 (avoid re-downloads), R-3 (disk usage control), R-4 (proactive feedback) — all covered.

No security, alignment, or constitution concerns found.

File: internal/cache/complypack.go [hunk 1]
```diff
@@ -130,9 +135,9 @@ func (c *ComplypackCache) Store(config complypack.Config, content io.Reader) (st
 		return "", fmt.Errorf("failed to write config file: %w", err)
 	}
 
-	// Evict old version directories for the same evaluator-id.
-	// This ensures at most one version exists per evaluator-id after Store().
-	evictOldVersions(parentDir, config.Version)
+	// Evict old version directories for the same evaluator-id, keeping
+	// up to N versions per the configured retention count (FR-001).
+	evictOldVersions(parentDir, config.Version, CacheRetentionCount(), c.state)
```
trevor-vaughan: This needs to be below the `Rename` doesn't it? Otherwise, we could have evicted the cache then had a failure that would leave us with an inconsistent state.

internal/cache/state.go
```diff
@@ -154,3 +154,27 @@ func (s *State) GetComplypackState(repository string) (PolicyState, bool) {
...
+// EvaluatorIDToVersion performs a reverse lookup on the Complypacks map,
+// returning the active version for the given evaluator-id. State is keyed
+// by repository, so this iterates all entries to find the matching
+// evaluator-id. Returns ("", false) when the evaluator-id is not found,
+// or when the receiver or Complypacks map is nil/empty.
+//
+// Assumes one repository per evaluator-id (current invariant enforced by
+// complyctl get's duplicate evaluator-id rejection). If multiple
+// repositories reference the same evaluator-id, the returned version is
+// non-deterministic (Go map iteration order). If multi-repository-per-
```
trevor-vaughan: We should really fail here instead of having a non-deterministic result.

internal/cache/complypack_sync.go
trevor-vaughan, line 71: This will bail early if the remote is not available and the added code will never be reached for the local cache lookup.

internal/cache/complypack_sync.go
```diff
@@ -143,3 +176,47 @@ func (s *ComplypackSync) SyncComplypack(ctx context.Context, repository, version
...
+		vr, verifyErr := s.verifier(ctx, registryRef)
+		if verifyErr != nil {
+			return false, verifyErr
+		}
+		verifyResult = vr
+	}
+
+	// Update state to reflect the version switch and persist.
+	s.state.UpdateComplypackStateWithVerification(
+		repository, version, remoteDigest,
```
trevor-vaughan, line 233: Do we want the local content `digest` here not the `remoteDigest`?

internal/cache/complypack.go
trevor-vaughan: We should call `CacheRetentionCount()` once and store it as a variable. It calls `os.Getenv` which means that we might have issues with tests that manipulate environment variables.

Alternatively, we could add an option to `NewComplypackCache` that would allow tests to inject a retention count instead of manipulating an environment variable.

cmd/complyctl/cli/providers.go
```diff
@@ -77,7 +77,7 @@ func (o *providersOptions) run(ctx context.Context) error {
 		return nil
 	}
 
-	cc := cache.NewComplypackCache(o.cacheDir)
+	cc := cache.NewComplypackCache(o.cacheDir, nil)
```
trevor-vaughan: Does passing `nil` here by default cause issues with the state-driven lookup?

internal/doctor/doctor.go
```diff
@@ -811,6 +812,135 @@ func unmappedReason(resolver PolicyGraphResolver, resolveFailures int) string {
...
+	Untracked   bool
+}
+
+// walkCacheSize sums file sizes under {cacheDir}/complypacks/ using
+// filepath.WalkDir. Returns 0 if the directory does not exist (FR-007).
+func walkCacheSize(cacheDir string) (int64, error) {
+	complypacksDir := filepath.Join(cacheDir, complytime.ComplypacksSubdir)
+
+	var totalBytes int64
+	err := filepath.WalkDir(complypacksDir, func(_ string, d fs.DirEntry, err error) error {
```
trevor-vaughan: This follows symlinks and could walk spaces that are outside of the cache.

If the space is fully managed by us, and will not contain symlinks, then we should not follow them.

We may want to allow the cache root to be a symlink.

trevor-vaughan (changes requested):

internal/cache/complypack.go
yvonnedevlinrh: Addressed in `0d785145`: moved `evictOldVersions()` after the successful `os.Rename()` so old versions are preserved if the rename fails.

internal/doctor/doctor.go
yvonnedevlinrh: Addressed in `0d785145`: `walkCacheSize()` now resolves the cache root via `filepath.EvalSymlinks` (allowing the root to be a symlink), then skips any symlinks encountered during the walk via `d.Type()&fs.ModeSymlink` check.

internal/cache/state.go
yvonnedevlinrh: Addressed in `0d785145`: `EvaluatorIDToVersion()` now detects duplicate evaluator-id entries and returns an error instead of a non-deterministic result. Added `TestEvaluatorIDToVersion_DuplicateEvaluatorID`.

internal/cache/complypack_sync.go
yvonnedevlinrh: Addressed in `0d785145`: restructured `SyncComplypack()` so that when `DefinitionVersion()` fails (registry unreachable), the code checks whether a local cache entry exists before returning an error. If cached data is available, it serves from local cache with a warning.

internal/cache/complypack_sync.go
yvonnedevlinrh: Keeping `remoteDigest` is intentional. State tracks the registry manifest digest so that the incremental sync check (`localState.Digest == remoteDigest`) can skip redundant fetches on subsequent runs. Using a local content digest would cause a permanent mismatch and force re-fetches every time.

internal/cache/complypack.go
yvonnedevlinrh: Addressed in `0d785145`: added `retentionCount` field to `ComplypackCache`, populated once at construction via `CacheRetentionCount()`. `Store()` now uses the cached field instead of calling `os.Getenv` on each invocation.

cmd/complyctl/cli/providers.go
yvonnedevlinrh: Addressed in `0d785145`: `providers.go` now loads state via `cache.LoadState()` and passes it to `NewComplypackCache`, following the same graceful degradation pattern as `doctor.go`. If state cannot be loaded, nil is passed and the directory-scan fallback applies.

trevor-vaughan (approved):

gvauter (approved):
Well-structured feature implementation covering all 10 functional requirements (FR-001 through FR-010) from the OpenSpec. Architecture is clean: retention count read once at construction, eviction uses state timestamps for ordering (orphaned first, then oldest), local cache hits avoid network round-trips with proper re-verification when sigstore is configured.

Graceful degradation throughout — nil state falls back to directory scan, corrupt state produces warnings rather than hard errors, registry unavailability serves from local cache. Test coverage is thorough with 15+ new test functions.

One minor observation (LOW, non-blocking): `CacheRetentionCount()` has no upper bound — a defensive cap (e.g., 100) with a warning could prevent unbounded disk growth from misconfiguration. Not blocking since the value is user-controlled via an explicit env var.

PR #711, opened 2026-07-10
feat: surface Gemara bundle metadata in list and get commands Body:

### LINKED ISSUE #506: feat: surface Gemara bundle metadata via CLI for informed configuration
## Problem

After `complyctl get` pulls a Gemara bundle from an OCI registry, the CLI discards everything it learned about the bundle content. The user has no way to discover:

- What policy is inside the bundle (its canonical name, title, description)
- What evaluator it requires (openscap, ampel, opa)
- What target variables that evaluator needs
- Whether their chosen `id` in `complytime.yaml` has any relationship to the actual policy

The `policies[].id` field in `complytime.yaml` is a purely local alias — it is never validated against or suggested from the bundle's internal metadata. A user writing `id: banana` would work identically to `id: cis-fedora-l1-workstation-policy`. The config has zero awareness of what's inside the bundle.

## What the bundle already contains

The OCI config blob (fetched during `get`) has structured metadata:

```json
{
  "bundle-version": "1",
  "artifacts": [
    {
      "id": "cis-fedora-l1-workstation-policy",
      "type": "Policy",
      "role": "artifact",
      "dependencies": ["cis-fedora-l1-workstation-catalog.yaml", "cis-fedora-l1-guidance.yaml"]
    }
  ]
}
```

The policy YAML layer adds: `title`, `metadata.id`, `adherence.evaluation-methods[].executor.id`, and the full assessment plan structure.

All of this is available after a minimal fetch (manifest + config blob = 2 HTTP requests) — the full layers don't even need to be pulled for basic discovery.

## Proposed approach

Introduce a way for users to discover bundle metadata before or after pulling. Options (not mutually exclusive):

### Option A: `complyctl inspect <url>`

Lightweight pre-pull discovery. Fetches only the manifest + config blob from the registry (no full pull). Prints:

```
$ complyctl inspect quay.io/test_complytime/complytime-policies@v0.3.0-rc1-grc

Policy:    cis-fedora-l1-workstation-policy
Title:     CIS Fedora Linux - Level 1 Workstation Policy
Evaluator: openscap
Requires:  profile, datastream
Artifacts: 3 (1 policy, 1 catalog, 1 guidance)
```

### Option B: Enrich `complyctl list` output

After `get`, show bundle-internal metadata alongside the user-chosen ID:

```
POLICY ID   VERSION         BUNDLE POLICY                        EVALUATOR
cis-fedora  v0.3.0-rc1-grc  cis-fedora-l1-workstation-policy    openscap
```

### Option C: `complyctl get` prints a summary

After successful sync, report what was pulled:

```
Syncing policy 1/1: cis-fedora... done
  → Policy: cis-fedora-l1-workstation-policy (openscap)
  → Required variables: profile
  → Artifacts: 3 layers (72K policy, 89K catalog, 12K guidance)
```

## Why this matters

- Users currently have to `oras blob fetch` and manually inspect JSON to understand what they pulled
- The `id` in `complytime.yaml` is disconnected from bundle identity — a config can look correct but reference the wrong policy version/content without any warning
- With multiple policies from the same repository (different tags), the derived ID would always be the same (`complytime-policies`), giving no hint about what's actually inside
- New providers (OPA, future evaluators) will make this worse — users need to know which evaluator a policy uses to know what variables to configure

## References

- `internal/policy/loader.go` — `isBundleManifest()`, `LoadBundleFiles()`, `DetectManifestShape()`
- `internal/cache/` — `NewSync()`, OCI layout access
- `internal/registry/client.go` — `NewRemoteRepository()`, manifest fetch
- Bundle manifest type: `bundle.MediaTypeManifest` (`application/vnd.gemara.manifest.v1+json`)

## Related

- #505 — doctor should validate target variables against provider requirements (complementary: that issue catches missing variables; this issue helps users know what variables to provide in the first place)

yvonnedevlinrh:
## Summary

Implements [#506](https://github.com/complytime/complyctl/issues/506) (Options B + C) -- surface Gemara bundle metadata via CLI for informed configuration.

### Changes

- **`complyctl list`** gains EVALUATOR and CONTROLS columns showing the evaluator type and control count for each cached policy
- **`complyctl get`** prints a post-sync summary to stderr after fetching a policy, showing title, evaluator, control count, and assessment count
- **`PolicyState`** in `state.json` gains `policy_title`, `policy_evaluator`, `control_count`, and `assessment_count` fields (backward compatible via `omitempty`)
- **`policy.Resolver.ExtractPolicyMetadata()`** provides lightweight metadata extraction without building a full `DependencyGraph`
- **Upgrade backfill** automatically populates metadata for pre-existing caches on the next `complyctl get` run
- **`formatPolicySummary()`** helper for testable stderr output

### Files Changed

| Area | Files |
|------|-------|
| CLI commands | `cmd/complyctl/cli/get.go`, `cmd/complyctl/cli/list.go` |
| Cache state | `internal/cache/state.go` |
| Policy resolver | `internal/policy/resolver.go` |
| Tests | `cli_test.go`, `get_verify_test.go`, `state_test.go`, `resolver_test.go` |
| Docs | `CHANGELOG.md`, `AGENTS.md` |
| OpenSpec | `openspec/changes/bundle-metadata-cli/` (proposal, design, spec, tasks) |

### Option A (deferred)

`complyctl inspect <url>` for pre-pull discovery was evaluated and deferred -- the gap is minimal now that `get` prints a summary and `list` shows metadata.

Closes #506


cmd/complyctl/cli/cli_test.go
```diff
@@ -576,13 +576,346 @@ func TestListOptions_Run_UncachedPolicyShowsDash(t *testing.T) {
...
+func TestListOptions_Run_ColumnOrder(t *testing.T) {
+	chdirTemp(t)
+	writeWorkspaceConfig(t, minimalConfig)
+
+	cacheDir := t.TempDir()
+	state := &cache.State{
+		Policies: map[string]cache.PolicyState{
+			"policies/test-policy": {
+				Version:         "v1.0",
+				Digest:          "sha256:9f86d081884c7d65",
```
trevor-vaughan: This is an invalid hash and (ideally) would result in a failure somewhere.

cmd/complyctl/cli/get.go
```diff
@@ -395,9 +406,87 @@ func syncSinglePolicy(
...
+	if !fetched {
+		ps, _ := state.GetPolicyState(ref.Repository)
+		if !hasMetadata(ps) {
+			extractMetadata = true
+		}
+	}
+
+	if extractMetadata {
+		meta, metaErr := resolver.ExtractPolicyMetadata(
+			ref.Repository, version,
```
trevor-vaughan: There's a subtle bug here in the way `:latest` is handled. `SyncPolicy` tags the store with the concrete registry version and persists it to `state.Policies[repo].Version`. The current code binds `latest` so we end up dropping metadata on retrieval (code sussing assisted by Claude).

```suggestion
   // hasMetadata() (not just PolicyEvaluator=="") avoids redundant
   // re-extraction for multi-evaluator policies where PolicyEvaluator
   // is intentionally empty.
   ps, _ := state.GetPolicyState(ref.Repository)
   extractMetadata := fetched || !hasMetadata(ps)
   
   if extractMetadata {
        // Use the concrete version SyncPolicy persisted, not the config
        // tag: a ":latest" pin is stored under the resolved version, so
        // extracting by "latest" would fail store.Resolve() and silently
        // drop all metadata.
        meta, metaErr := resolver.ExtractPolicyMetadata(
            ref.Repository, ps.Version,
       ) 
```

cmd/complyctl/cli/get.go
```diff
@@ -395,9 +406,87 @@ func syncSinglePolicy(
...
+			)
+		} else {
+			state.SetPolicyMetadata(
+				ref.Repository,
+				meta.Title,
+				meta.EvaluatorID,
+				meta.ControlCount,
+				meta.AssessmentCount,
+			)
+			if saveErr := cache.SaveState(
```
trevor-vaughan: This should be moved to the end of the loop so that we aren't writing to disk repeatedly and instead batch the flush.

cmd/complyctl/cli/get.go
```diff
@@ -299,16 +301,23 @@ func syncAllPolicies(
 	policies []complytime.PolicyEntry,
 	wsCfg *complytime.VerificationConfig,
 	vfCache map[complytime.VerificationConfig]cache.VerifyFunc,
+	cacheDir string,
```
trevor-vaughan: Do we need this as opposed to using `cacheMgr.Dir()`?

cmd/complyctl/cli/cli_test.go
```diff
@@ -576,13 +576,346 @@ func TestListOptions_Run_UncachedPolicyShowsDash(t *testing.T) {
...
+		Common:   &Common{Output: Output{Out: &buf}},
+		cacheDir: cacheDir,
+	}
+	err := o.run(context.Background())
+	require.NoError(t, err)
+
+	output := buf.String()
+	// Controls should show "25" since metadata exists.
+	assert.Contains(t, output, "25")
+	// The data row should show "-" for evaluator (multi-evaluator).
```
trevor-vaughan, line 823: This test doesn't actually check for  a dash in the output.

trevor-vaughan (changes requested):

cmd/complyctl/cli/cli_test.go
yvonnedevlinrh: Addressed in b6cdfcb5: replaced truncated SHA-256 digests with valid-length 64 hex char hashes across all test fixtures added by this PR.

cmd/complyctl/cli/get.go
yvonnedevlinrh: Addressed in b6cdfcb5: applied the suggested fix — extractMetadata now uses fetched || !hasMetadata(ps) and passes ps.Version (the concrete persisted version) to ExtractPolicyMetadata instead of the config tag.

cmd/complyctl/cli/get.go
yvonnedevlinrh: Addressed in b6cdfcb5: moved cache.SaveState() from syncSinglePolicy to after the loop in syncAllPolicies for a single batch flush.

cmd/complyctl/cli/get.go
yvonnedevlinrh: addressed in b6cdfcb5: removed cacheDir parameter from syncAllPolicies and syncSinglePolicy; using cacheMgr.Dir() instead.

cmd/complyctl/cli/cli_test.go
yvonnedevlinrh: Addressed in b6cdfcb5: added assertion that parses the EVALUATOR column position from the header and verifies the data row shows "-" for multi-evaluator policies.

yvonnedevlinrh:
@trevor-vaughan Thanks for the thorough review. All five items addressed in b6cdfcb5:
1. Invalid test digests — replaced truncated SHA-256 hashes with valid-length 64 hex char digests across test fixtures.
2. :latest version bug — applied your suggestion: extractMetadata uses fetched || !hasMetadata(ps) and passes ps.Version (concrete persisted version) to ExtractPolicyMetadata.
3. Batch SaveState — moved cache.SaveState() from per-policy in syncSinglePolicy to a single flush after the loop in syncAllPolicies.
4. Redundant cacheDir param — removed from syncAllPolicies/syncSinglePolicy; using cacheMgr.Dir() instead.
5. Missing dash assertion — TestListOptions_Run_MultiEvaluatorShowsDash now asserts the evaluator column contains "-".
Also removed omitempty from ControlCount/AssessmentCount JSON tags so zero values survive round-trips — no impact on existing caches since json.Unmarshal produces 0 for missing int fields either way. Branch rebased onto current main to resolve the merge conflict.

trevor-vaughan (approved):

marcusburghardt (approved):
LGTM. I saw a minor detail regarding the errors messages. I will create a follow-up PR on top of it.

PR #708, opened 2026-07-09
feat: add emoji status indicators to markdown report for CI step summary

em-redhat:
## Summary

Add emoji status indicators to the `--format pretty` Markdown compliance
report so it can be used directly as a GitHub Actions Step Summary,
replacing ~55 lines of custom Python in org-infra's
`reusable_compliance.yml`.

### Changes

- **`resultEmoji()` helper** (`internal/output/markdown.go`): Maps
  `gemara.Result` to `complytime.Status*` emoji constants
  (`Passed`→✅, `Failed`→❌, `NotApplicable`/`NotRun`→⏭️,
  `Unknown`/`NeedsReview`→⚠️)
- **Controls table**: Each control and requirement row prefixed with
  emoji in the Result column
- **Summary counts headers**: Emoji-prefixed column headers
  (`✅ Passed | ❌ Failed | ⚠️ Needs Review | ...`)
- **Findings group headers**: Emoji-prefixed section headers
  (`### ❌ Failed`, `### ⚠️ Needs Review`, etc.)

### Testing

- Table-driven `TestResultEmoji` unit test (7 cases including default)
- 3 new integration tests with section-scoped assertions:
  `TestMarkdown_ControlsTableStatusIndicators`,
  `TestMarkdown_SummaryCountsTableHeaders`,
  `TestMarkdown_FindingsGroupHeaderEmoji`
- 6 existing tests updated for emoji presence
- All 76 tests pass in `internal/output`

### Downstream

Enables `complytime/org-infra` to simplify CI by replacing custom
Python summary generation with:
```sh
cat report-*.md >> $GITHUB_STEP_SUMMARY
```

### Spec Artifacts

OpenSpec change: `openspec/changes/ci-step-summary-report/`

Closes #697

PR #725, opened 2026-07-14
fix: add text labels alongside emoji in terminal scan summary

### LINKED ISSUE #724: fix: add text labels alongside emoji in terminal scan summary STATUS column
## Summary

The terminal scan summary (`complyctl scan`) renders the STATUS column with emoji-only values (`✅`, `❌`, `⏭️`, `⚠️`). Screen readers, braille displays, and monochrome terminals may skip, garble, or fail to distinguish these characters, making the status effectively invisible for visually impaired users.

The markdown report (PR #708) already uses `emoji + text` (e.g., `✅ Passed`). The terminal output should follow the same pattern for consistency and accessibility.

## Current behavior

```
TARGET ID             REQUIREMENT ID   CONTROL ID              STATUS  MESSAGE
complytime-complyctl  block-force-push force-push-restriction   ✅      Force pushing is disabled
```

## Expected behavior

```
TARGET ID             REQUIREMENT ID   CONTROL ID              STATUS     MESSAGE
complytime-complyctl  block-force-push force-push-restriction   ✅ Passed  Force pushing is disabled
```

## Proposed fix

One-line change in `internal/output/scan_summary.go:154`:

```go
// Before:
rows = append(rows, []string{e.targetID, e.requirementID, e.controlID, e.emoji, e.message})

// After:
rows = append(rows, []string{e.targetID, e.requirementID, e.controlID, e.emoji + " " + e.result.String(), e.message})
```

Tests in `scan_summary_test.go` will need updating to expect the text label alongside the emoji.

## Context

Identified during review of PR #708 (emoji status indicators in markdown report).

marcusburghardt:
## Summary

Add text labels alongside emoji in the terminal scan summary STATUS column for accessibility. Screen readers, braille displays, and monochrome terminals may skip or garble emoji-only values, making the status effectively invisible for visually impaired users.

The STATUS column now displays `✅ Passed` instead of `✅`, matching the pattern already used in the markdown report (`--format pretty`).

### Changes

- **`internal/output/scan_summary.go`**: Append `result.String()` to the emoji in the STATUS column cell (`e.emoji + " " + e.result.String()`)
- **`internal/output/scan_summary_test.go`**: Update `TestFormatScanSummary_PassingWithEmptyMessage` assertion to expect the result text as the last field

### Before

```
STATUS
✅
❌
⏭️
⚠️
```

### After

```
STATUS
✅ Passed
❌ Failed
⏭️ Not Applicable
⚠️ Unknown
```

## Related Issues

- Closes #724

## Review Hints

- Single-commit PR with a one-line production code change and a test update. Review as a single commit.
- The change mirrors how the markdown report already formats status indicators (PR #708).
- `terminal.ShowPlainTable` auto-sizes columns, so the wider STATUS values adjust automatically.

### DIFF
--- a/internal/output/scan_summary.go
+++ b/internal/output/scan_summary.go
@@ -151,7 +151,7 @@ func FormatScanSummary(assessments []provider.AssessmentLog, assessmentTargets [
 	headers := []string{"TARGET ID", "REQUIREMENT ID", "CONTROL ID", "STATUS", "MESSAGE"}
 	var rows [][]string
 	for _, e := range entries {
-		rows = append(rows, []string{e.targetID, e.requirementID, e.controlID, e.emoji, e.message})
+		rows = append(rows, []string{e.targetID, e.requirementID, e.controlID, e.emoji + " " + e.result.String(), e.message})
 	}
 
 	conclusion := fmt.Sprintf("%d requirements: %d passed, %d failed, %d not applicable, %d skipped, %d errors",
--- a/internal/output/scan_summary_test.go
+++ b/internal/output/scan_summary_test.go
@@ -370,12 +370,12 @@ func TestFormatScanSummary_PassingWithEmptyMessage(t *testing.T) {
 	lines := strings.Split(output, "\n")
 	for _, line := range lines {
 		if strings.Contains(line, complytime.StatusPassed) {
-			// After the status emoji, only whitespace should remain (empty message)
+			// After the status text, only whitespace should remain (empty message)
 			parts := strings.Fields(line)
-			// Fields: target1, REQ-1, CTRL-1, emoji — no message field
+			// Fields: target1, REQ-1, CTRL-1, emoji, result text — no message field
 			lastField := parts[len(parts)-1]
-			assert.Equal(t, complytime.StatusPassed, lastField,
-				"StatusPassed emoji should be the last non-empty field when message is empty")
+			assert.Equal(t, "Passed", lastField,
+				"result text should be the last non-empty field when message is empty")
 		}
 	}
 }


em-redhat (dismissed):
LGTM. Great to see a focus on accessibility

trevor-vaughan (dismissed):

marcusburghardt:
Hi Trevor, thanks for looking at this and for the intent behind adding broader test coverage for the result text in the STATUS column.

Before merging, I'd like to share some observations about the assertions introduced in `f076afa7`.

**One assertion is demonstrably weak.** The `SortOrder` test asserts `Contains(output, "Failed")`, but this matches the MESSAGE column (`Message: "Failed"` at line 311), not the STATUS column. You can verify this by reverting the production change and observing the assertion still passes:

```bash
# 1. Confirm tests pass with the feature
GOTOOLCHAIN=auto go test -race -count=1 ./internal/output/... -run "SortOrder$" -v

# 2. Revert only the production change (keep the tests)
sed -i 's/e.emoji + " " + e.result.String()/e.emoji/' internal/output/scan_summary.go

# 3. Run again -- "Failed" assertion still passes (matches MESSAGE, not STATUS)
GOTOOLCHAIN=auto go test -race -count=1 ./internal/output/... -run "SortOrder$" -v

# 4. Restore the production change
sed -i 's/e.emoji,/e.emoji + " " + e.result.String(),/' internal/output/scan_summary.go
```

In step 3, `"Unknown"`, `"Not Run"`, and `"Passed"` correctly fail -- but `"Failed"` passes because it matches the step message, not the STATUS column. A test that passes regardless of whether the feature exists doesn't validate the feature.

The remaining five assertions do catch the revert, but only because the test data happens not to contain those title-case strings elsewhere. They're structurally imprecise -- `Contains` on the full output can't distinguish which column produced the match.

**This was a deliberate scoping decision.** The existing scan summary tests don't assert on result text in the STATUS column for other result types either. If we want to add that coverage, it should be done consistently across all scan summary tests in a dedicated PR, rather than partially within this one. This keeps the change atomic and avoids inconsistency across similar test cases. Additionally, the value of these assertions is questionable -- as demonstrated above, `Contains` on the full output can't reliably distinguish which column produced the match, and the individual components are already verified at the function level: the emoji mapping is covered by existing `StatusPassed`/`StatusFailed`/`StatusError` assertions in the scan summary tests, and the result text (`"Failed"`, `"Passed"`, etc.) is provided by `gemara.Result.String()` in the upstream library. Asserting on their concatenation in the formatted output couples tests to presentation details without adding meaningful regression protection.

I also replaced the merge commit with a rebase to keep the PR history linear and consistent with the project conventions.

Pushing commits to open PRs is welcome for minor fixes and confirmed cases, but a quick confirmation before pushing would help avoid situations like this. A short comment like "I'd like to add assertions for the other result types -- does that align with the scope?" would have surfaced the deliberate scoping decision and the assertion pattern concerns upfront, saving effort on both sides.

I'll drop this commit and keep the PR as a single-commit change. Happy to collaborate on a follow-up PR that adds comprehensive STATUS column assertions across all scan summary tests with a stronger pattern (e.g., row-scoped checks using `emoji + " " + text` compound strings).

trevor-vaughan (approved):

em-redhat (approved):

PR #727, opened 2026-07-14
fix(policy): warn when split-layer catalog load fails

marcusburghardt:
## Summary

`extractSplitMetadata` in `internal/policy/resolver.go` silently swallowed catalog layer load errors, defaulting `ControlCount` to `0` without any indication to the user. This made it impossible to distinguish "policy has no catalog layer" from "catalog layer exists but failed to load" (e.g., post-sync disk corruption).

This PR adds a `log.Warn` when `LoadLayerByMediaType` fails for the catalog layer, following the warning pattern already used in `internal/cache/`. The behavior is unchanged — `ControlCount` defaults to `0` and no error is returned — but the user now sees a warning in the log output.

This addresses the catalog error handling asymmetry between `extractBundleMetadata` (which reports catalog issues via the files map) and `extractSplitMetadata` (which previously swallowed them). Identified during review of #711.

## Related Issues

- Follow-up to #711

## Review Hints

- This is a 2-file, 12-line change. The only behavioral difference is a new `log.Warn` — return values are unchanged.
- The existing `TestResolver_ExtractPolicyMetadata_SplitMissingCatalog` test covers the return-value contract (ControlCount=0, no error). The warning log is not directly tested, consistent with how `internal/cache/` handles its `log.Warn` calls (none are tested for output).
- The `charmbracelet/log` import is new to `internal/policy/` but follows the established pattern in `internal/cache/retention.go` and `internal/cache/complypack.go`.

### DIFF
--- a/internal/policy/resolver.go
+++ b/internal/policy/resolver.go
@@ -5,6 +5,8 @@ package policy
 import (
 	"fmt"
 
+	"github.com/charmbracelet/log"
+
 	"github.com/complytime/complyctl/internal/complytime"
 	"github.com/gemaraproj/go-gemara"
 	"github.com/goccy/go-yaml"
@@ -207,7 +209,14 @@ func (r *Resolver) extractSplitMetadata(
 
 	catalogData, catalogLoadErr := r.loader.LoadLayerByMediaType(
 		policyID, version, complytime.MediaTypeCatalog)
-	if catalogLoadErr == nil {
+	if catalogLoadErr != nil {
+		// Catalog layer unavailable — could be genuinely absent or
+		// corrupted after sync. Log a warning so the user knows the
+		// control count may be inaccurate rather than silently
+		// defaulting to zero.
+		log.Warn("catalog layer unavailable, control count will be 0",
+			"policy", policyID, "error", catalogLoadErr)
+	} else {
 		catalog, catErr := parseControlCatalog(catalogData)
 		if catErr != nil {
 			return PolicyMetadata{}, fmt.Errorf(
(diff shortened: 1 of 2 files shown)


internal/policy/resolver.go
```diff
@@ -207,7 +209,14 @@ func (r *Resolver) extractSplitMetadata(
 
 	catalogData, catalogLoadErr := r.loader.LoadLayerByMediaType(
 		policyID, version, complytime.MediaTypeCatalog)
-	if catalogLoadErr == nil {
+	if catalogLoadErr != nil {
+		// Catalog layer unavailable — could be genuinely absent or
+		// corrupted after sync. Log a warning so the user knows the
+		// control count may be inaccurate rather than silently
+		// defaulting to zero.
+		log.Warn("catalog layer unavailable, control count will be 0",
```
yvonnedevlinrh, line 217: Low finding : The log.Warn includes "policy" and "error" structured fields but omits "version". The LoadLayerByMediaType call on line 210 uses both policyID and version. If a user has multiple versions of the same policy cached, the warning won't distinguish which version has the missing catalog.

yvonnedevlinrh (dismissed):
LGTM. Only low finding, nothing to block this merge

internal/policy/resolver.go
marcusburghardt: Addressed in `b2a88eaf`: added "version" structured field to the log.Warn call so users can distinguish which version has the unavailable catalog.

gvauter (approved):
Clean, well-scoped fix. The `log.Warn` follows the established pattern from `internal/cache/` (9 existing call sites), import grouping is correct, return-value contract is unchanged, and existing test coverage is sufficient.

No issues found.

yvonnedevlinrh (approved):
LGTM

PR #789, opened 2026-07-29
feat: centralize report filename construction

### LINKED ISSUE #773: OSCAL/SARIF scan output overwritten per-target — only last target's results survive
## Bug

`complyctl scan -f oscal` with multiple `targets` configured only records findings for the **last** target in the output file — every other target's results are silently lost.

## Root cause

`ToOSCAL` (`internal/output/oscal.go:95`) names its output file from policy ID + second-granularity timestamp only, no target ID:

```go
filename := fmt.Sprintf("assessment-results-%s-%s.json",
    complytime.FilenameSafe(policyID), time.Now().Format("20060102-150405"))
```

`processScanOutput` (`cmd/complyctl/cli/scan.go`) writes one report per target in a tight loop, so all targets resolve to the same filename within the same second and each overwrites the last. `ToSARIF` (`internal/output/sarif.go:33`) has the same bug.

By contrast, `Evaluator.Write` (`internal/output/evaluator.go:155`), which writes the per-target YAML evaluation log, *does* include target ID in its filename — which is why those files correctly appear one-per-target while OSCAL/SARIF don't.

## Repro

Configure a policy against N>1 targets, run `complyctl scan -p <policy> -f oscal`. `.complytime/scan/` ends up with only one `assessment-results-*.json`, and every finding in it references only the last target — confirmed on a 23-target run where all findings read `"target soar-4151: ..."` (the last target in config) despite the log showing all 23 targets scanned and 23 separate per-target `evaluation-log-*.yaml` files present.

## Expected

Each target's OSCAL/SARIF report should get its own file (include target ID in filename, matching evaluation-log), or be merged into one combined multi-target document.

## Impact

Silent data loss — scan exits 0, summary looks fine, but the OSCAL artifact (the one most likely used as compliance evidence) only reflects one target.

trevor-vaughan:
## Summary

Multi-target scans under the same policy produced OSCAL, SARIF, and Markdown reports with identical filenames (only `policyID` + timestamp), causing overwrites when targets processed within the same second. The EvaluationLog formatter already included `targetID` and was unaffected.

This PR adds a `BuildReportFilename` helper in `internal/output/filename.go` that all four formatters now call. Report filenames include the target ID, matching the EvaluationLog pattern.

## Related Issues

- Fixes #773

## Changes

- **New helper:** `BuildReportFilename(prefix, policyID, targetID, ext)` in `internal/output/filename.go`
  - Sanitizes IDs via `FilenameSafe`, timestamps via `time.Now().Format("20060102-150405")`
  - Omits targetID segment when empty (no double dashes)
- **Formatters migrated:** OSCAL, SARIF, Markdown, and EvaluationLog now call the shared helper
  - Removes inline `fmt.Sprintf` from each formatter
- **Tests updated:**
  - `filename_test.go` covers sanitization, empty targetID, compound extensions
  - Formatter tests verify targetID appears in output filenames
  - `TestMultiTargetDistinctFilenames` regression test for #773
- **Docs:** CHANGELOG.md and AGENTS.md entries added

## Review Hints

- The filename helper is in `internal/output/filename.go` with comprehensive tests in `filename_test.go`
- Each formatter (`evaluator.go`, `oscal.go`, `sarif.go`, `markdown.go`) has a one-line change replacing `fmt.Sprintf` with `BuildReportFilename`
- The regression test `TestMultiTargetDistinctFilenames` in `oscal_test.go` reproduces the original bug (scanning two targets produces two distinct files)
- All OpenSpec artifacts are in `openspec/changes/multi-target-report-filenames/` (proposal, design, spec, tasks)

## Testing

- [x] Unit tests pass (`make test-unit`)
- [x] Lint passes (`make lint`)
- [x] Vet passes (`make vet`)
- [x] CRAP load check passes (`make crapload-check`)
- [x] Regression test added (`TestMultiTargetDistinctFilenames`)


SecKatie (approved):
Seems like a good fix!

em-redhat (dismissed):
Clean, well-tested fix for multi-target report overwrites (#773). `BuildReportFilename` is a solid centralization, import cleanup is thorough, and test coverage is strong across all formatters. One non-blocking nit on `openspec/config.yaml` noted inline.

openspec/config.yaml
```diff
@@ -0,0 +1 @@
+schema: spec-driven
```
em-redhat, line 1: nit: This file already exists on `main` with 20 lines of scaffolding comments (context, rules examples, etc.). The diff shows it as a new file with just `schema: spec-driven`, which means the branch deleted and re-created it, stripping the template content. Looks unintentional — consider reverting this file to its `main` state.

trevor-vaughan: Hi @em-redhat, it looks like that one was deleted in 5c311436d71c82b5bf8ee798b7b783366298b1fb. It's auto-created by the openspec process so, honestly, I just treat it as noise :-/.

em-redhat: Makes sense, by no means a blocker anyways. Approval stands

hbraswelrh (dismissed):
LGTM. Nice change.

hbraswelrh (approved):
LGTM.

em-redhat (approved):

PR #831, opened 2026-08-27
fix: do not create workspace directory as a side effect of logging

marcusburghardt:
## Problem

When running any `complyctl` command from `$HOME` without a prior `complyctl init`, the `lazyLogWriter` creates `~/.complytime/` as a side effect of writing the log file. On subsequent runs, `CheckLegacyDir()` detects this directory and emits a false deprecation warning about migrating to XDG directories — even though no legacy data exists.

### Reproduction

```bash
# From a clean home directory (no ~/.complytime)
cd ~
complyctl list
# First run: creates ~/.complytime/ via error logging, shows only the error
# Second run: CheckLegacyDir() finds ~/.complytime/, shows false WARNING + error
```

### Root cause

`lazyLogWriter.Write()` in `cmd/complyctl/cli/root.go` called `os.MkdirAll()` to create `{workspace}/.complytime/` before writing the log file. When `workspace == $HOME`, this path (`~/.complytime`) collides with the legacy directory that `CheckLegacyDir()` warns about.

## Fix

Replace `os.MkdirAll` with `os.Stat` — the log writer now only writes when the workspace directory **already exists** (i.e., the user previously ran `complyctl init`). When the directory does not exist, log output is silently discarded.

This is safe because:
- Commands fail early without a workspace config, and the error is already printed to stderr
- Debug mode (`--debug`) tees all output to stderr via `teeWriter` regardless of log file creation
- In normal operation, `complyctl init` creates the workspace directory before any meaningful logging occurs

## Changes

| File | Change |
|---|---|
| `cmd/complyctl/cli/root.go` | Replace `os.MkdirAll` with `os.Stat` check in `lazyLogWriter.Write()` |
| `cmd/complyctl/cli/cli_test.go` | Update existing tests to pre-create workspace dir; add `TestLazyLogWriter_Write_SkipsWhenDirNotExists`; update `TestEnableDebug_UnwritableLogDir` → `TestEnableDebug_NonexistentLogDir` |

## Testing

```bash
go test -race ./...   # all pass
go vet ./...          # clean
```

### DIFF
--- a/cmd/complyctl/cli/cli_test.go
+++ b/cmd/complyctl/cli/cli_test.go
@@ -1756,6 +1756,9 @@ func TestSaveGenerationAndPrint_NoComplypacks(t *testing.T) {
 
 func TestLazyLogWriter_Write_CreatesLogFile(t *testing.T) {
 	baseDir := t.TempDir()
+	// Pre-create the workspace directory to simulate an initialized workspace.
+	require.NoError(t, os.MkdirAll(filepath.Join(baseDir, complytime.WorkspaceDir), 0700))
+
 	w := &lazyLogWriter{}
 	w.SetWorkspace(baseDir)
 
@@ -1774,6 +1777,9 @@ func TestLazyLogWriter_Write_CreatesLogFile(t *testing.T) {
 
 func TestLazyLogWriter_Write_DefaultBaseDir(t *testing.T) {
 	chdirTemp(t)
+	// Pre-create the workspace directory to simulate an initialized workspace.
+	require.NoError(t, os.MkdirAll(complytime.WorkspaceDir, 0700))
+
 	w := &lazyLogWriter{}
 	// No SetWorkspace call — should default to "."
 
@@ -1797,6 +1803,9 @@ func TestLazyLogWriter_Close_NilFile(t *testing.T) {
 
 func TestLazyLogWriter_Write_MultipleWrites(t *testing.T) {
 	baseDir := t.TempDir()
+	// Pre-create the workspace directory to simulate an initialized workspace.
+	require.NoError(t, os.MkdirAll(filepath.Join(baseDir, complytime.WorkspaceDir), 0700))
+
 	w := &lazyLogWriter{}
 	w.SetWorkspace(baseDir)
 
@@ -1811,6 +1820,26 @@ func TestLazyLogWriter_Write_MultipleWrites(t *testing.T) {
 	content, err := os.ReadFile(logPath)
 	require.NoError(t, err)
 	assert.Equal(t, "first\nsecond\n", string(content))
+}
+
+func TestLazyLogWriter_Write_SkipsWhenDirNotExists(t *testing.T) {
+	baseDir := t.TempDir()
+	// Do NOT pre-create the workspace directory.
+	// The writer must not create it as a side effect.
+	w := &lazyLogWriter{}
+	w.SetWorkspace(baseDir)
+
+	n, err := w.Write([]byte("should be discarded\n"))
+
+	require.NoError(t, err)
+	assert.Equal(t, 20, n)
+
+	wsDir := filepath.Join(baseDir, complytime.WorkspaceDir)
+	_, statErr := os.Stat(wsDir)
+	assert.True(t, os.IsNotExist(statErr),
+		"workspace directory must not be created as a side effect of logging")
+
+	require.NoError(t, w.Close())
 }
 
 // --- New() tests ---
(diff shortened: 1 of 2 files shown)


em-redhat (approved):
Clean  fix. The logic is good. Logging is never skipped in a scenario where the user would lose visibility into errors. The teeWriter design ensures debug output always reaches stderr, and main.go independently prints errors to stderr. The test coverage is thorough, including both the positive and negative case.

cmd/complyctl/cli/root.go
```diff
@@ -54,8 +54,11 @@ func (w *lazyLogWriter) Write(p []byte) (int, error) {
 			baseDir = "."
 		}
 		logDir := filepath.Join(baseDir, complytime.WorkspaceDir)
-		if err := os.MkdirAll(logDir, 0700); err != nil {
-			fmt.Fprintf(os.Stderr, "Warning: failed to create log directory: %v\n", err)
+		// Only write logs when the workspace directory already exists.
+		// Do not create it as a side effect of logging — creating
+		// ~/.complytime when workspace is $HOME would trigger a false
+		// legacy-directory deprecation warning from CheckLegacyDir().
+		if _, err := os.Stat(logDir); err != nil {
```
yvonnedevlinrh, line 61: **[MEDIUM] `os.Stat` branch collapses all error classes into a silent skip**

`if _, err := os.Stat(logDir); err != nil { return }` treats a permission
error (EACCES) or any other non-`IsNotExist` failure identically to "dir
does not exist," discarding logs with no diagnostic — where the pre-fix code
emitted `Warning: failed to create log directory`. Suggest narrowing the
silent skip to the not-exist case and warning on genuine failures, which
keeps the fix's intent (no dir creation, no false legacy warning) while
restoring diagnostics:

```suggestion
		// Only write logs when the workspace directory already exists.
		// Do not create it as a side effect of logging — creating
		// ~/.complytime when workspace is $HOME would trigger a false
		// legacy-directory deprecation warning from CheckLegacyDir().
		if _, err := os.Stat(logDir); err != nil {
			if !os.IsNotExist(err) {
				fmt.Fprintf(os.Stderr, "Warning: failed to stat log directory: %v\n", err)
			}
			return
		}

File: cmd/complyctl/cli/cli_test.go [hunk 2]
```diff
@@ -1813,6 +1822,26 @@ func TestLazyLogWriter_Write_MultipleWrites(t *testing.T) {
 	assert.Equal(t, "first\nsecond\n", string(content))
 }
 
+func TestLazyLogWriter_Write_SkipsWhenDirNotExists(t *testing.T) {
```
yvonnedevlinrh, line 1825: **[MEDIUM] Lost coverage of the filesystem-error branch**

Repurposing `TestEnableDebug_UnwritableLogDir` → `TestEnableDebug_NonexistentLogDir`
removed the only test that exercised a filesystem-error path through
`lazyLogWriter`. The `Warning: failed to open log file` branch (root.go:67)
is now unreached by the suite. Consider adding a case where the workspace
dir exists but the log file cannot be opened (e.g., pre-create
`.complytime/complyctl.log` as a read-only or as a directory) and assert the
warning fires. If Comment 1's `IsNotExist` narrowing is adopted, also add a
stat-error case.

yvonnedevlinrh (approved):
_This review was generated by /uf.review-pr (AI-assisted)._

**Verdict: APPROVE** (non-blocking suggestions)

A small, well-targeted bugfix. Replacing `os.MkdirAll` with an `os.Stat` gate in `lazyLogWriter.Write()` correctly stops `~/.complytime` from being created as a side effect of logging, which was triggering the false legacy-directory deprecation warning from `CheckLegacyDir()`.

Verified:
- **Correctness**: `io.Writer` contract still honored (returns `len(p), nil` on the discard path); all byte-count test data is valid (14 / 12 / 20 bytes).
- **Caller impact**: return semantics unchanged from the pre-fix `MkdirAll`-failure path; the `TeeWriter` consumer ignores the secondary writer's return, and `PersistentPreRun`'s `Debug log:` hint does not depend on the file existing. No caller regressions.
- **Reachability**: both `os.Stat` branches are reachable and (for the skip/create paths) tested. No dead branches introduced.

Two related non-blocking observations, sharing one root cause (the`os.Stat` branch collapses all error classes into a silent skip):

1. **[MEDIUM] Silent swallow of non-`IsNotExist` errors** — a permission error (EACCES) on an existing-but-inaccessible dir is now discarded with no diagnostic, where the old code warned. Narrow the silent skip to `os.IsNotExist(err)` and warn otherwise.
2. **[MEDIUM] Lost test coverage** — repurposing `TestEnableDebug_UnwritableLogDir` → `TestEnableDebug_NonexistentLogDir` removed the only test exercising a filesystem-error branch; the `Warning: failed to open log file` path (root.go:67) is now unreached by the suite.

And one minor doc-fidelity nit:

**[LOW / pre-existing] Doc now stale as a side effect of this change**

Not touched by this diff, but this change makes the existing comment
inaccurate: creation is no longer guaranteed on write — it's now conditional
on the workspace dir existing. Since this PR is what changes that behavior,
updating the doc here would be reasonable, though it's optional and could be
a follow-up.

None block merge; the PR is correct for its primary path and already carries
an APPROVE from @em-redhat.

PR #840, opened 2026-09-04
feat: evidence source provenance (EvidenceMapping) support

### LINKED ISSUE #841: feat: support Evidence source provenance (EvidenceMapping) in provider API
## Problem

The Gemara `#Evidence` schema includes a `source` field of type
`#EvidenceMapping` ([evaluationlog.cue#L73](https://github.com/gemaraproj/gemara/blob/e74ec8a1f2bb81262dd643972602eca4c34348ad/evaluationlog.cue#L73),
[auditlog.cue#L78-97](https://github.com/gemaraproj/gemara/blob/e74ec8a1f2bb81262dd643972602eca4c34348ad/auditlog.cue#L78))
that records **where** evidence was collected from -- which artifact,
which location within it, and a content digest for integrity pinning.

The go-gemara SDK already exposes this as `Evidence.Source` (`EvidenceMapping`
struct with fields `ReferenceId`, `Coordinate`, `EntryId`, `Digest`,
`Remarks`).

complyctl's gRPC provider API (`api/plugin/plugin.proto`) currently
carries only the basic evidence fields (`id`, `type`, `description`,
`payload`, `collected_at`) and explicitly defers the `source` field
with a comment referencing an unwritten "ADR 0023".

Without `source`, providers cannot report evidence provenance --
auditors and downstream consumers cannot trace an evidence entry back
to the specific artifact, file path, API endpoint, or Gemara entry
that produced it.

## Proposed Solution

- Add `EvidenceMapping` proto message with fields `reference_id`,
  `coordinate`, `entry_id`, `digest`, `remarks`
- Add `source` field (type `EvidenceMapping`, field 6) to the
  existing `Evidence` proto message (backward-compatible)
- Add `EvidenceSource` struct to `pkg/provider/client.go`
- Plumb through proto mapping functions, evaluator, and Markdown
  formatter
- Update test provider for E2E coverage

## References

- Draft PR with OpenSpec artifacts: #840
- Provider-side adoption: tracked separately in complytime-providers
- Gemara schema: `#Evidence` in `auditlog.cue`, `#EvidenceMapping` in `mapping_inline.cue`

marcusburghardt:
## Summary

Plumb the Gemara `#EvidenceMapping` (`source` field) through complyctl's
entire evidence pipeline, enabling providers to report **where** evidence
was collected from.

Fixes #841

## Context

The Gemara CUE schema (`evaluationlog.cue` line 73, `auditlog.cue`
lines 78-97) defines `#Evidence` with a `source` field of type
`#EvidenceMapping` that records evidence provenance -- which artifact,
which location within it, and a content digest. The go-gemara SDK already
exposes this as `Evidence.Source`. complyctl's proto API currently defers
this field with a comment referencing an unwritten "ADR 0023".

## Changes

- **Proto API**: Add `EvidenceMapping` message (`reference_id`,
  `coordinate`, `entry_id`, `digest`, `remarks`) and `source` field
  (field 6) on `Evidence`. Remove stale ADR 0023 comment.
- **Provider SDK**: Add `EvidenceSource` struct and
  `Source *EvidenceSource` field on `Evidence` in `pkg/provider/`.
- **Proto mapping**: `internalEvidenceSourceToProto()` in `server.go`
  and `protoEvidenceMappingToInternal()` in `client.go`.
- **Evaluator**: Map `provider.EvidenceSource` to
  `gemara.EvidenceMapping` when constructing evaluation log entries.
- **Markdown**: Render source provenance as `source: <ref> @ <coord>`
  in `formatEvidenceMeta()`.
- **Test provider**: Emit evidence with `Source` for E2E coverage.
- **Tests**: Comprehensive coverage for both proto mapping directions,
  evaluator serialization (YAML + JSON), and markdown rendering.

Backward-compatible: providers that don't send `source` are unaffected.

## OpenSpec Artifacts

- `openspec/changes/evidence-source/proposal.md`
- `openspec/changes/evidence-source/specs/evidence-source/spec.md`
- `openspec/changes/evidence-source/design.md`
- `openspec/changes/evidence-source/tasks.md`

## Related

- Gemara schema reference: https://github.com/gemaraproj/gemara/blob/e74ec8a1f2bb81262dd643972602eca4c34348ad/evaluationlog.cue#L73
- Provider-side adoption: [complytime/complytime-providers#174](https://github.com/complytime/complytime-providers/issues/174)
- Upstream evidence-to-OSCAL/SARIF conversion: [gemaraproj/go-gemara#127](https://github.com/gemaraproj/go-gemara/issues/127)
- OSCAL formatter migration: [#851](https://github.com/complytime/complyctl/issues/851)


openspec/changes/evidence-source/design.md
```diff
@@ -0,0 +1,116 @@
...
+
+- Full alignment with go-gemara `Evidence.Source` field
+- Backward-compatible proto change (additive field)
+- Round-trip fidelity for all five `EvidenceMapping` subfields
+- Evidence source rendered in Markdown reports
+- Evidence source serialized in YAML/JSON evaluation logs
+
+**Non-Goals:**
+
+- OSCAL output evidence support (separate concern, no upstream
```
jpower432: In OSCAL Assessment Results there is an `relevant_evidence` field under observations. If we are using the upstream conversion (we should be), I submitted an issue to add support https://github.com/gemaraproj/go-gemara/issues/127

marcusburghardt: Great @jpower432 . I updated the this section to include more context. This doesn't change much the current change proposal, but I agree we should adopt upstream conversion. So I create this new issue https://github.com/complytime/complyctl/issues/851 to address this once we are good to move completely to  `gemaraconv`.

gxmiranda (dismissed):
Clean, well-specified feature implementation that plumbs the Gemara `EvidenceMapping` through complyctl's entire evidence pipeline -- from proto definition through gRPC mapping, evaluator, and Markdown rendering. Comprehensive test coverage (both mapping directions, nil/partial/full cases, YAML/JSON serialization), backward-compatible design, and proper spec-first workflow with OpenSpec artifacts. All CI checks pass.

api/plugin/plugin.proto
```diff
@@ -142,9 +142,33 @@ message Step {
...
 
+// EvidenceMapping identifies the artifact or system from which
+// evidence was collected. Mirrors go-gemara EvidenceMapping.
+message EvidenceMapping {
+  // Ties this evidence to a mapping-reference in the artifact's metadata
+  string reference_id = 1;
+
+  // Precise location within the referenced artifact (e.g. file path,
+  // API path, JSON path). Upstream Gemara schema documents this as
+  // mutually exclusive with entry_id; complyctl passes through without
```
jpower432: This was a documentation bug. Not mutally exclusive. Fixed in the next release.

api/plugin/plugin.proto
```diff
@@ -142,9 +142,33 @@ message Step {
   string message = 3;
 }
 
+// EvidenceMapping identifies the artifact or system from which
+// evidence was collected. Mirrors go-gemara EvidenceMapping.
+message EvidenceMapping {
+  // Ties this evidence to a mapping-reference in the artifact's metadata
+  string reference_id = 1;
```
jpower432, line 149: Do plugins have awareness of a mapping reference? Would we prefer that they send us a list of their sources too?

jpower432 (changes requested):
Implementation looks good, but I see an issue with regard to `reference-id`. There doesn't seem to be a way to collect this actual `reference` from plugins and I don't think the host would have that information either.

api/plugin/plugin.proto
marcusburghardt: Addressed in `cd214893`: removed the mutual exclusivity language from the proto comments on `coordinate` and `entry_id`, as well as from the design doc (D5) and spec scenario. Thanks for the correction.

api/plugin/plugin.proto
marcusburghardt: Great question. I explored this in depth, including running through some possible scenarios with AI assistance against existing policy content in the repo.

**Current state**: This PR implements the transport layer — `reference_id` flows through from provider → proto → evaluator → output. If the provider sets it, it appears in the evaluation log and Markdown report. If not, it's empty.

**The gap you identified**: Providers may not know about `mapping-references` from the policy metadata. The host (complyctl) has this information (via the `DependencyGraph`), but currently doesn't pass it to the evaluator.

**Host-side enrichment** would require:
1. Preserve `MappingReferences` from `parsePolicyLayer()` in the resolver (currently discarded)
2. Build a `reqToMappingRef` map (requirement → catalog mapping-reference id)
3. Pass it to the `Evaluator` and enrich evidence that has empty `reference_id`
4. Populate `EvaluationLog.Metadata.MappingReferences` in the output (currently empty) — without this, `reference_id` values would be dangling references

**Trade-offs**:
- The matching heuristic (requirement → catalog → mapping-reference) works for the common case but `reference_id` is generic — evidence could relate to guidance or threats, not just catalogs
- Both enrichment AND metadata output must ship together for the reference to be meaningful
- It touches the resolver (core pipeline), which needs its own design consideration

**Options**:
1. **Keep this PR as transport** — providers can populate `reference_id` if they have the info; follow up with host-side enrichment in a dedicated PR with proper spec
2. **Expand this PR** — add host-side enrichment now (scope expansion: resolver + evaluator + metadata output changes)

I'm leaning toward option 1 (transport now, enrichment follow-up) to keep this PR focused, but I'd value your input on whether you think the enrichment should block this PR or can be a fast follow-up.

marcusburghardt: Addressed in `dae3b509`: added MappingReferences propagation from policy metadata through the resolver pipeline to `EvaluationLog.Metadata.MappingReferences`.

**Investigation findings:**

You were right that `reference_id` had a structural gap. We traced the full data flow and found:

1. `gemara.EvidenceMapping.ReferenceId` is a pointer into `Metadata.MappingReferences` -- the EvaluationLog must declare the references for the cross-reference to resolve.
2. The policy resolver (`extractFromGemaraPolicy`) was discarding `p.Metadata.MappingReferences` during extraction.
3. The evaluator (`GemaraLog`) was creating `Metadata{}` with no `MappingReferences`, so any `reference_id` value was a dangling reference.
4. OSCAL BackMatter (via `gemaraconv.EvaluationLogToOSCALAssessmentResults`) was silently empty because this data never reached the EvaluationLog.

**What this commit adds:**

- `DependencyGraph.MappingReferences` -- extracted from policy metadata
- `NewEvaluator` gains a `mappingReferences` parameter
- `GemaraLog()` populates `Metadata.MappingReferences` in the output
- Scan pipeline wires `graph.MappingReferences` through to evaluator
- Tests cover extraction (split/bundle), propagation, YAML/JSON serialization, and nil/omitempty behavior

**What remains (follow-up):**

Host-side enrichment of empty `reference_id` values on evidence. Currently, providers must populate `reference_id` themselves. A future enhancement could auto-enrich empty `reference_id` by matching requirements to their catalog mapping-references in the resolver. This is a separate concern that touches the resolver's core pipeline and deserves its own spec.

The other four `EvidenceMapping` fields (`coordinate`, `entry_id`, `digest`, `remarks`) work independently without `reference_id`, providing standalone evidence provenance value for provider adoption (complytime-providers#174).

marcusburghardt:
It should be ready for review again, but I still need to conclude some local tests by tomorrow.

marcusburghardt:
@jpower432 and @gxmiranda , I would appreciate another round of reviews. I addressed the previous concerns and did some tests to ensure no breaking changes and smooth adoption by providers. There is also a [follow-up issue](https://github.com/complytime/complyctl/issues/876) focusing on documentation. You may benefit of the examples in [this comment](https://github.com/complytime/complyctl/issues/876#issuecomment-5773893091) when reviewing the PR. Thanks!

gxmiranda (approved):
Clean, well-structured feature that plumbs Gemara `EvidenceMapping` through complyctl's entire evidence pipeline (proto → SDK → evaluator → output). Includes MappingReferences propagation from policy metadata addressing review feedback. Comprehensive test coverage across all conversion boundaries (bidirectional proto mapping, evaluator serialization in YAML/JSON, Markdown rendering). Backward-compatible — providers that don't send `source` are unaffected. All 31 CI checks pass.

_This review was generated by /uf.review-pr (AI-assisted)._

api/plugin/plugin.proto
jpower432: Apologies for missing this comment earlier @marcusburghardt. I realize my feedback might have been ambiguous about the direction of propagation. I would expect the providers to be providing mapping references information to the host. The `Evidence.reference-id` represents references that pertain to the source of evaluation inputs. Any example might be the a raw payload from hitting the GitHub API. The solution you have here would work if the those evidence source URL are available as part of the Policy metadata, but I don't know that they realistically would be.

marcusburghardt: Thanks for the clarification, @jpower432.
I see now where we misaligned. After checking once more the artifacts in `go-gemara` I can confirm your concern about providers declaring runtime data sources is valid. I created the #879, which includes the full analysis, schema references, and implementation plan. In that issue I named our concerns in "Direction A" and "Direction B", both valid based on Gemara docs I found.

This PR addresses Direction A (policy metadata → EvaluationLog). Direction B is complementary, not redundant. The Gemara canonical fixture [good-evaluation-log.yaml](https://github.com/gemaraproj/go-gemara/blob/main/test-data/good-evaluation-log.yaml) shows both types coexisting. Policy catalog refs alongside runtime source refs. Without Direction A, `control.reference-id` and `requirement.reference-id` dangle on every scan.

Proposal: merge as-is to avoid scope creeping in this PR. Direction A code is reused without rework when #879 lands. WDYT?

jpower432 (approved):
LGTM

Issue #765, opened 2026-07-23
chore(doctor): address pre-existing tech debt found during PR #722 review


sonupreetam:
## Context

During the `/review-council` for PR #722 (`doctor-grouped-output`), the Divisor agents flagged several pre-existing issues in `internal/doctor/` and `cmd/complyctl/cli/doctor.go` that were **not introduced** by the PR but surfaced during review.

### 1. `registryErrors` cache bypassed by `resolvePinnedFallback` (SRE HIGH)

**File**: `internal/doctor/doctor.go`

The `registryErrors` cache (line 370) prevents redundant `ResolveLatestVersion` calls to unreachable registries. However, when a cached error is found, `resolvePinnedFallback` still calls `resolver.ResolveVersion()` -- making a new network call to the same unreachable registry.

With N pinned policies from one unreachable registry: the first policy triggers 2 timeouts (latest + pinned), each subsequent triggers 1 timeout (pinned fallback). For 5 policies at 5s timeout = 30s of wasted waits.

**Fix**: Extend the `registryErrors` cache to be consulted inside `resolvePinnedFallback`, or skip the pinned resolution attempt when the error is a known network failure (not a 404).

### 2. `CheckVariables` complexity 38, 211 lines (Architect MEDIUM)

**File**: `internal/doctor/doctor.go:580-791`

`CheckVariables` handles four responsibilities: evaluator-target resolution, global variable validation, target variable validation, and verbose detail generation. Gaze flags it as Q4 Dangerous (complexity 38, GazeCRAP 38).

**Fix**: Extract `resolveEvaluatorTargets()` and `validateProviderVariables()` helpers. Each would be under 50 lines (CS-010).

### 3. `Run()` has 8 parameters (Architect MEDIUM)

**File**: `internal/doctor/doctor.go:117`

The signature exceeds the 99-character line limit and violates AP-001 (Options struct pattern).

**Fix**: Introduce a `RunOptions` struct to bundle the parameters.

### 4. `printDiagnostics` writes to stdout, not `io.Writer` (Architect LOW)

**File**: `cmd/complyctl/cli/doctor.go:108-167`

Uses `fmt.Println`/`fmt.Printf` directly instead of accepting `io.Writer` per AP-003 (testable CLI pattern). This will be naturally addressed when #611 (machine-readable output) is implemented.

**Fix**: Accept `io.Writer` parameter, use `fmt.Fprintln`/`fmt.Fprintf`. Likely addressed alongside #611.

## Priority

[...]

PR #846, opened 2026-09-08
fix(doctor): skip pinned resolution for unreachable registries

marcusburghardt:
## Summary

Addresses #765 item 1 (SRE HIGH): `registryErrors` cache bypassed by
`resolvePinnedFallback`.

## Problem

The `registryErrors` cache in `CheckPolicyVersions` prevents redundant
`ResolveLatestVersion` calls to unreachable registries. However,
`resolvePinnedFallback` still called `ResolveVersion` against the same
unreachable registry for each pinned policy, triggering one additional
timeout per policy.

For N pinned policies from one unreachable registry at 5s timeout: the
first policy triggers 2 timeouts (latest + pinned), each subsequent
triggers 1 timeout (pinned fallback). For 5 policies = up to 30s of
wasted waits in offline/air-gapped environments.

## Fix

Guard the `ResolveVersion` call in `resolvePinnedFallback` with a check
on whether the error indicates the registry is reachable:

- **Network error** (not `ErrVersionNotFound`): registry is unreachable,
  skip pinned resolution entirely
- **404** (`ErrVersionNotFound`): registry IS reachable but the latest
  tag is missing, proceed with pinned version resolution as before

This eliminates all redundant timeout waits for unreachable registries.

## Changes

- **`internal/doctor/doctor.go`**: Add `isRegistryReachable` check before
  `ResolveVersion` call in `resolvePinnedFallback`
- **`internal/doctor/doctor_test.go`**: Add `resolveVersionCalls` counter
  to mock, add two tests verifying the optimization (unreachable skips,
  reachable 404 proceeds)

## Verification

- `make test-unit` -- all tests pass
- `make lint` -- zero issues

Ref #765

### DIFF
--- a/internal/doctor/doctor.go
+++ b/internal/doctor/doctor.go
@@ -553,6 +553,10 @@ func CheckPolicyVersions(cfg *complytime.WorkspaceConfig, dataDir string, versio
 // resolvePinnedFallback attempts to resolve a pinned version when the latest
 // tag is unavailable. Returns a pass result if the pinned version resolves,
 // or a warn result with a user-friendly diagnosis.
+//
+// When latestErr is a network error (not ErrVersionNotFound), the pinned
+// resolution is skipped because the registry is unreachable — attempting
+// ResolveVersion would trigger another timeout against the same host.
 func resolvePinnedFallback(
 	resolver VersionResolver,
 	ref complytime.PolicyRef,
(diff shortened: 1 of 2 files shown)


em-redhat (approved):
Minimal, correct fix with thorough tests. The logic is consistent with the existing error classification model in the caller. All CI checks pass.

sonupreetam (approved):
LGTM

